PR #2, opened 2021-05-26
modify the poc code to use velero discovery helper

alaypatel07:
Creating a Draft PR,

I have a bunch of todos to work on before I can call this ready for review

### DIFF
--- a/cmd/export/discover.go
+++ b/cmd/export/discover.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/pager"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -158,7 +160,7 @@ func resourceToExtract(namespace string, dynamicClient dynamic.Interface, lists 
 				APIResource:     resource,
 			}
 
-			objs, err := getObjects(g, namespace, dynamicClient)
+			objs, err := getObjects(g, namespace, dynamicClient, log)
 			if err != nil {
 				switch {
 				case apierrors.IsForbidden(err):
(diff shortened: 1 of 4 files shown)


shawn-hurley (commented):
I assume a todo is to use a logger, I notice that we are already bringing in logrus, is that because that is what kustomize or something uses? would it be usefull to use the same logger?

sseago:
@shawn-hurley Velero uses logrus for all of its logging, so that may have been the reason for pulling it in here.

alaypatel07:
yes, logrus due to Velero.

One todo that is not in your list is to use pager[1] for the list call to avoid churn on apiserver/etcd 

1. https://github.com/kubernetes/client-go/blob/master/tools/pager/pager.go

cmd/export/discover.go
```diff
@@ -0,0 +1,205 @@
...
+	APIGroup        string
+	APIVersion      string
+	APIGroupVersion string
+	APIResource     metav1.APIResource
+	objects         *unstructured.UnstructuredList
+}
+
+type groupResourceError struct {
+	APIResource metav1.APIResource `json:",inline"`
+	Err         error              `json:"error"`
```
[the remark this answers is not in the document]
alaypatel07: updated

cmd/export/discover.go
```diff
@@ -0,0 +1,205 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
```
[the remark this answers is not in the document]
alaypatel07: updated as per slack conversation

cmd/export/discover.go
```diff
@@ -0,0 +1,214 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// groupResource contains the APIGroup and APIResource
+type groupResource struct {
+	APIGroup        string
+	APIVersion      string
+	APIGroupVersion string
+	APIResource     metav1.APIResource
+	objects         *unstructured.UnstructuredList
```
shawn-hurley, line 27: nit: either make all the fields unexported or non. either way the type is unexported so I don't see a need to mix it in this type.

cmd/export/export.go
```diff
@@ -0,0 +1,188 @@
...
+	if o.Namespace == "" {
+		log.Debugf("--namespace is empty, defaulting to current context namespace\n")
+		if currentContext.Namespace == "" {
+			log.Errorf("current context %s namespace is empty, exiting\n", contextName)
+			return fmt.Errorf("current context %s namespace is empty, exiting\n", contextName)
+		}
+		o.Namespace = currentContext.Namespace
+	}
+
+	log.Debugf("current context is: %s\n", currentContext.AuthInfo)
```
shawn-hurley, line 120: https://pkg.go.dev/k8s.io/client-go@v0.21.1/tools/clientcmd/api#AuthInfo

I think that we maybe want to use say user/namespace/cluster here just so we don't leak passwords/tokens etc.

cmd/export/export.go
```diff
@@ -0,0 +1,188 @@
...
+
+	restConfig, err := o.configFlags.ToRESTConfig()
+	if err != nil {
+		log.Errorf("cannot create rest config: %#v", err)
+		return err
+	}
+
+	dynamicClient := dynamic.NewForConfigOrDie(restConfig)
+
+	features.NewFeatureFlagSet()
```
shawn-hurley, line 157: Why do we need to do this? is this because of the velero changes we need? do we have an issue we can add in a comment here?

shawn-hurley (approved):
Issues for follow up

1. Debug switch we probably want to consider log levels here.
2. Flag for specifying the kubeconfig

besides some nite
/lgtm

cmd/export/export.go
alaypatel07: +1 velero, I have to create the upstream issue and follow up on it.

cmd/export/export.go
alaypatel07: will fix this asap

cmd/export/discover.go
alaypatel07: +1, will follow up

PR #5, opened 2021-06-15
Apply CLI

shawn-hurley:
Implementation of the apply CLI that will enable running transformation plugins.

Notable Changes:
* Created internal file helpers to consolidate file manipulation across subcommands.
* PathOpts helps get all the path and file names that are needed.


cmd/apply/apply.go
```diff
@@ -0,0 +1,138 @@
...
+	cmd.Flags().StringVarP(&o.ExportDir, "transform-dir", "t", "transfrom", "The path where files that contain the transformations are saved")
+	cmd.Flags().StringVarP(&o.ExportDir, "output-dir", "o", "output", "The path where files are to be saved after transformation are applied")
+}
+
+func (o *Options) run() error {
+	// log := o.logger
+	a := apply.Applier{}
+
+	// Load all the resources from the export dir
+	p, err := filepath.Abs(o.ExportDir)
```
alaypatel07, line 73: nit: name the var as `exportDir` instead of p?

cmd/apply/apply.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		// Handle errors better for users.
+		return err
+	}
+
+	t, err := filepath.Abs(o.TransformDir)
```
alaypatel07, line 79: nit: `transformDir` var?

internal/file/file_helper.go
```diff
@@ -0,0 +1,92 @@
...
+			})
+		}
+	}
+	return jsonFiles, nil
+}
+
+//TODO: @shawn-hurley Add errors for these methods to validate that the correct struct values are set.
+type PathOpts struct {
+	TransformDir string
+	ResourceDir  string
```
alaypatel07, line 69: nit: name this `ExportDir` for consistency?

cmd/apply/apply.go
```diff
@@ -0,0 +1,138 @@
...
+		TransformDir: t,
+		ResourceDir:  p,
+		OutputDir:    outputDir,
+	}
+
+	for _, f := range files {
+		whPath := opts.GetWhiteOutFilePath(f.Info.Name())
+		_, statErr := os.Stat(whPath)
+		if !os.IsNotExist(statErr) {
+			o.logger.Debugf("resource file: %v is skipped due to white file: %v", f.Info.Name(), whPath)
```
alaypatel07, line 104: this one feels like `Info` log level to me? WDYT

internal/file/file_helper.go
```diff
@@ -0,0 +1,92 @@
...
+			if err != nil {
+				return nil, err
+			}
+			json, err := yaml.YAMLToJSON(data)
+			if err != nil {
+				return nil, err
+			}
+
+			u := unstructured.Unstructured{}
+			u.UnmarshalJSON(json)
```
shawn-hurley, line 55: There could be an error here. need to capture and check it

internal/file/file_helper.go
```diff
@@ -0,0 +1,92 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/yaml"
+)
+
+type File struct {
+	Info         os.FileInfo
+	Unstructured unstructured.Unstructured
+}
+
+func ReadFiles(dir string) ([]File, error) {
```
shawn-hurley, line 20: Add a context, we will want to make this timeout potentially

cmd/apply/apply.go
```diff
@@ -0,0 +1,138 @@
...
+			continue
+		}
+
```
alaypatel07, line 107: nit: add a comment here each resource needs to have a transform or a whiteout

shawn-hurley: It is expected that this be the case, probably should either error or handle that huh

cmd/apply/apply.go
shawn-hurley: I went back and forth for now probably makes sense because we can't do debug log level yet

internal/file/file_helper.go
```diff
@@ -0,0 +1,92 @@
...
+	log := logrus.New()
+
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, err
+	}
+	return readFiles(dir, files, log)
+}
+
+func readFiles(path string, files []os.FileInfo, log *logrus.Logger) ([]File, error) {
```
alaypatel07, line 30: I wonder if we would want to make this a walker eventually. Not totally sure though

alaypatel07 (approved):
Added some naming nits, overall makes sense to me.

/lgtm

internal/file/file_helper.go
shawn-hurley: Maybe the recursive thing works fine for now I think, what is the benefit of a walker?

cmd/apply/apply.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	addFlagsForOptions(o, cmd)
+
+	return cmd
+}
+
+func addFlagsForOptions(o *Options, cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&o.ExportDir, "export-dir", "e", "export", "The path where the kubernetes resources are saved")
+	cmd.Flags().StringVarP(&o.ExportDir, "transform-dir", "t", "transfrom", "The path where files that contain the transformations are saved")
```
sseago, line 64: copy/paste error? `&o.ExportDir` should be `&o.TransformDir`?

cmd/apply/apply.go
```diff
@@ -0,0 +1,138 @@
...
+	cmd.Flags().StringVarP(&o.ExportDir, "output-dir", "o", "output", "The path where files are to be saved after transformation are applied")
```
sseago, line 65: copy/paste error? `&o.ExportDir` should be `&o.OutputDir`?

cmd/apply/apply.go
sseago: Also, s/transfrom/transform/

cmd/apply/apply.go
shawn-hurley: Added a comment for a to-do item to handle this.

internal/file/file_helper.go
shawn-hurley: I agree, this might be an optimization later but it works now so I think I we should just keep it and can fix in a min.

cmd/apply/apply.go
```diff
@@ -0,0 +1,140 @@
...
+		TransformDir: transformDir,
+		ExportDir:    exportDir,
+		OutputDir:    outputDir,
+	}
+
+	//TODO: @shawn-hurley handle case where transform or whiteout file is not present.
+	for _, f := range files {
+		whPath := opts.GetWhiteOutFilePath(f.Info.Name())
+		_, statErr := os.Stat(whPath)
+		if !os.IsNotExist(statErr) {
```
alaypatel07, line 105: nit: I wonder if this can give side effects that might bite us

alaypatel07 (approved):
/lgtm for merging 

Added a small nit but can do a follow on if it is something that needs to be handled

cmd/apply/apply.go
```diff
@@ -0,0 +1,140 @@
...
+
+	opts := file.PathOpts{
+		TransformDir: transformDir,
+		ExportDir:    exportDir,
+		OutputDir:    outputDir,
+	}
+
+	//TODO: @shawn-hurley handle case where transform or whiteout file is not present.
+	for _, f := range files {
+		whPath := opts.GetWhiteOutFilePath(f.Info.Name())
```
sseago, line 103: s/f.Info.Name()/f.Path/

cmd/apply/apply.go
```diff
@@ -0,0 +1,140 @@
...
+		_, statErr := os.Stat(whPath)
+		if !os.IsNotExist(statErr) {
+			o.logger.Infof("resource file: %v is skipped due to white file: %v", f.Info.Name(), whPath)
+			continue
+		}
+
+		tfPath := opts.GetTransformPath(f.Info.Name())
```
sseago, line 110: s/f.Info.Name()/f.Path/

PR #8, opened 2021-06-18
initial implementation of transform command

sseago:


cmd/transform/transform.go
```diff
@@ -0,0 +1,162 @@
...
+	runner := transform.Runner{}
+
+	// Load all the resources from the export dir
+	exportDir, err := filepath.Abs(o.ExportDir)
+	if err != nil {
+		// Handle errors better for users.
+		return err
+	}
+
+	pluginDir, err := filepath.Abs(o.PluginDir)
```
shawn-hurley, line 80: What if we don't want to run any plugins?

cmd/transform/transform.go
```diff
@@ -0,0 +1,162 @@
...
+					return err
+				}
+				whFile.Close()
+			}
+			continue
+		} else {
+			// if whiteout file exists from prior run, remove it
+			whPath := opts.GetWhiteOutFilePath(f.Path)
+			_, statErr := os.Stat(whPath)
+			if !os.IsNotExist(statErr) {
```
shawn-hurley, line 133: What it would look like to instead overwrite the entire transform dir every time. 

I think that doing this complicated logic to try and keep old changes and such is going to be pretty error-prone.

cmd/transform/transform.go
sseago: If the plugin dir contains no plugins (or doesn't exist), then there will be no plugins processed. The result is that the transforms dir will contain all zero-length transforms (and there will be no whiteout files). Then again, if you intend to run no plugins, wouldn't simply not running transform be equivalent to this? Just skip transform and run apply. If apply doesn't find any whiteout or transform files, then the un-transformed resource files should end up as-is in the output dir, right?

cmd/transform/transform.go
sseago: Are you suggesting that we recursively delete the transform dir and its contents prior to writing any files?

sseago: But to be clear, the logic here does not attempt to keep old changes. The intent is to make the whiteout and patch files look like they should for the current run. For whiteout, the two questions are 1) does the file exist, and 2) should the file exist? If these two questions have different answers, then either create or delete the file as appropriate. Arguably, we could skip the "does it exist? question when we want it to exist and just always issue the create call. That would avoid having to check for prior existence in the whiteout case. For haveWhiteOut==false, though, I think we still need to check for existence and only delete if it exists from before. In the case of the transform patches file, it's simpler. We always write a file (even if it's zero length), so we always just open the file with `os.Create`, meaning we always get a new file, overwriting any prior files. The only difference in outcome if we nuke the top-level transform dir ahead of time is that if there a re files there that do not correspond to resources in the export dir, those will also be deleted. I'm assuming that apply will ignore these files if they aren't deleted anyway, so this probably wouldn't have any effect on the outcome.

cmd/transform/transform.go
shawn-hurley: I was wondering if this should error at that point? 

I am also thinking about the future a little bit, but that is not fleshed out. I think this makes sense for now.

sseago: Ahh. Yeah, we could eventually change this to error out here, but I'm thinking for now that if there are no plugins, just generating the empty patch files would be less disruptive on the cli workflows. But we can revisit as needed. For one thing, I have no idea how non-binary plugins will be configured once we add support there. The binary plugin dir may well be empty in cases where the plugins of interest are of other types.

cmd/transform/transform.go
shawn-hurley: They may not have an effect on the outcome, but I am wondering if that is a bad user expectation, I think for now if we do the TODO's that you outlined that this is acceptable and once we talk the next step with transform's then we will flush these out

sseago:
Minor update to the commit to remove a commented-out debug line that I forgot to get rid of.

cooktheryan:
Not sure if this matters or not but I was running the creation of crane from a container and hit
```
go: github.com/sirupsen/logrus@v1.6.0: missing go.sum entry; to add it:
454
	go mod download github.com/sirupsen/logrus```

PR #11, opened 2021-06-23
update transform code to use new runner library function

shawn-hurley:
Depends on https://github.com/konveyor/crane-lib/pull/20


sseago (approved):
I tested this with a couple of example plugins and it seems to work just like before the update.

PR #14, opened 2021-07-01
initial implementation of plugin priorities addition to transform cli

sseago:


cmd/transform/transform.go
```diff
@@ -87,6 +90,14 @@ func (o *Options) run() error {
 		return err
 	}
 
+	var ignoredPatchesDir string
+	if len(o.IgnoredPatchesDir) > 0 {
```
shawn-hurley, line 94: minor minor nit: I usually use would use `o.IgnoredPatchesDir != ""`. Don't know if it is that big of a deal just pointing it out.

internal/file/file_helper.go
```diff
@@ -71,22 +71,38 @@ func readFiles(ctx context.Context, path string, files []os.FileInfo, log *logru
...
+}
+
+func (opts *PathOpts) updateIgnoredPatchesDirPath(prefix, filePath string) string {
+	if len(opts.IgnoredPatchesDir) == 0 {
+		return ""
+	}
+	return opts.updatePath(opts.IgnoredPatchesDir, prefix, filePath)
+}
+
+func (opts *PathOpts) updatePath(updateDir, prefix, filePath string) string {
```
shawn-hurley, line 103: +1 on this refactor

shawn-hurley (approved):
This is looking really good!

cmd/transform/transform.go
sseago: I'm happy to change. I've used both in the past (golang's lack of a nil value for strings is annoying), and I think I actually use `!=0` more often. I can change this instance.

sseago: changed it.

PR #18, opened 2021-07-14
Handle missing transform files by creating the resource file

### LINKED ISSUE #15: apply should treat missing transform files as if a no-op file was found
Currently apply will skip creation of a resource's output file if a transform file is missing. A missing transform file indicates that either transform was never run, or a transform (or whiteout) file was deleted post-transform. If there is no whiteout or transform file for a resource, then the result should be equivalent to the case where a transform file exists but contains no operations (i.e. "[]" as its content).

shawn-hurley:
in the output directory

fixes #15

### DIFF
--- a/cmd/apply/apply.go
+++ b/cmd/apply/apply.go
@@ -108,24 +108,32 @@ func (o *Options) run() error {
 			continue
 		}
 
+		// Set doc to the object, only update the file if the transfrom file exists
+		doc, err := f.Unstructured.MarshalJSON()
+		if err != nil {
+			return err
+		}
+
 		tfPath := opts.GetTransformPath(f.Path)
 		// Check if transform file exists
 		// If the transform does not exist, assume that the resource file is
 		// not needed and ignore for now.
 		_, tfStatErr := os.Stat(tfPath)
-		if errors.Is(tfStatErr, os.ErrNotExist) {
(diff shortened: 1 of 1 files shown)


cmd/apply/apply.go
```diff
@@ -113,19 +113,28 @@ func (o *Options) run() error {
...
+			// Some other error here err out
+			return err
 		}
 
-		transformfile, err := os.ReadFile(tfPath)
-		if err != nil {
+		// Set doc to the object, only update the file if the transfrom file exists
+		doc, err := f.Unstructured.MarshalJSON()
+		if err != nil && !errors.Is(tfStatErr, os.ErrNotExist) {
+			// Some other error here err out
```
sseago, line 124: Shouldn't this always return err when err != nil? If err != nil, this means `MarshalJSON()` failed,  even if `tfPath` doesn't exist. In the case when err != nil and we're not returning err now (because the transform file doesn't exist), then later when we call JSONToYAML, that will fail, won't it? Or is that fine since we'll just return the err from that call? (although "can't transform nil input" may be less useful to return than whatever would have been returned here.

sseago (commented):
Looks good overall, although I had a concern about the error return conditional in one case.

cmd/apply/apply.go
shawn-hurley: I think that I am doing all the right things this diff is a little confusing though

sseago (approved):

PR #19, opened 2021-07-21
DEMO: add transfer-pvc subcommand to migrate pvc data

alaypatel07:
This PR adds a subcommand to transfer a single PVC from one cluster to another.

A sample invocation of this command is as follows:

```
crane transfer-pvc --pvc-name foo --source-context=source-ctx --destination-context=destination-ctx
```

In this sample, you need to have the source-context and destination-context pointing to the cluster and namespace of the pvc Foo. 

1. It is assumed that the source namespace will be equal to the destination namespace, this will change in the future.
2. The subcommand does not clean-up the resources it creates. In order to clean up manually before running use the following command:
```
$ oc --context=destination-ctx -n pvc-transfer-source delete pod,cm,secret,svc,route -l app=crane2 && oc --context=destination-ctx -n pvc-transfer-source delete pvc foo && oc --context=source-ctx -n pvc-transfer-source delete pod,cm,secret,svc,route -l app=crane2
```
3. Currently the tool assumes PVC does not exist at the destination. Eventually we might want to provide configuration options to users on how to handle this case


shawn-hurley:
LGTM, create the issues for follow up and then lets merge

PR #23, opened 2021-08-06
Add skip plugins functionality, and list-plugins subcommand

sseago:
Fixes https://github.com/konveyor/crane/issues/21


internal/plugin/plugin_helper.go
```diff
@@ -47,3 +48,31 @@ func getBinaryPlugins(path string, files []os.FileInfo) ([]transform.Plugin, err
 func isExecAny(mode os.FileMode) bool {
     return mode&0111 != 0
 }
+
+func GetFilteredPlugins(pluginDir, skipPlugins string) ([]transform.Plugin, error) {
+	var filteredPlugins []transform.Plugin
+	plugins, err := GetBinaryPlugins(pluginDir)
```
shawn-hurley, line 54: I think this is going to require an update. I think that we should consider GetFilteredPlugins moving to GetPlugins.

Also, can `skipplugins` be a vardic? and the caller responsible for the `split` ?

shawn-hurley (changes requested):
I think that we have a small merge conflict (sorry merged that one before this one)

sseago:
@shawn-hurley no worries on the merge conflict. I was expecting it -- and yes, we need to make sure that KubernetesPlugin can be skipped too. I'll work out the update.

internal/plugin/plugin_helper.go
sseago: @shawn-hurley I expect that GetFilteredPlugins will call GetPlugins now instead of GetBinaryPlugins, as we need to be able to skip internal plugins as well.

sseago: As for vardic, the skipPlugins string comes from the user. it's a comma-separated list, and this method splits it out. It's a CLI option, so it's a string.

sseago: I think in an earlier iteration, the caller was responsible for the split, but the result was that I had 3 identical split calls to pass into this func, since all 3 callers are different transform subcommands, so it was less code to just split it in here.

shawn-hurley: Fair enough, I realized after the fact that the API was internal so makes sense. 

The only other time is if that was to ever be used by something else maybe we want to do this in like validate flags or something?

shawn-hurley (approved):

internal/plugin/plugin_helper.go
sseago: Makes sense. If a second use for this comes up, it should be pulled out of here into a util method of some sort.

Issue #22, opened 2021-07-30
Enable internal k8s plugin for transforms by default


shawn-hurley:

shawn-hurley:
Fixed by #24

PR #24, opened 2021-08-09
Enable k8s plugins for transform

eemcmullan:
Enable k8s plugins from crane-lib for transform. 
#22

### DIFF
--- a/cmd/transform/optionals/optionals.go
+++ b/cmd/transform/optionals/optionals.go
@@ -10,8 +10,8 @@ import (
 )
 
 type Options struct {
-	logger            logrus.FieldLogger
-	PluginDir         string
+	logger    logrus.FieldLogger
+	PluginDir string
 }
 
 func (o *Options) Complete(c *cobra.Command, args []string) error {
@@ -65,7 +65,7 @@ func (o *Options) run() error {
 		return err
 	}
 
-	plugins, err := plugin.GetBinaryPlugins(pluginDir)
+	plugins, err := plugin.GetPlugins(pluginDir)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


sseago (changes requested):
This looks good, but I'd suggest a minor refactoring. Currently we have the concept of both binary and internal plugins (although we only have one example of the latter right now). GetBinaryPlugins pulls from the plugin dir. I think I'd rather preserve the functionality of GetBinaryPlugins as it exists today. Perhaps a new func `GetPlugins(dir string)` which starts with the kubernetes plugin (a list with one item in it), and then appends the results of `GetBinaryPlugins` to that list. Basically putting the new code from this PR into this new func, which calls the existing func. Otherwise people who weren't involved in this refactoring might be confused that `GetBinaryPlugins` returns both binary and internal plugins.

sseago:
Actually, with this refactoring, `GetBinaryPlugins` doesn't need to be an exported function, so forget what I said above. Leave it as-is, but just rename the exported func `GetPlugins` (and change the places where it's called).

eemcmullan:
Just updated `GetBinaryPlugins` to `GetPlugins`.

sseago (approved):

cmd/transform/optionals/optionals.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 type Options struct {
-	logger            logrus.FieldLogger
-	PluginDir         string
+	logger    logrus.FieldLogger
+	PluginDir string
```
shawn-hurley, line 14: +1 thanks for fixing gofmt

shawn-hurley (approved):
Looks good to me +1

thanks!

PR #30, opened 2021-08-27
Adding options to copy accross namespaces

shawn-hurley:
Depends on https://github.com/konveyor/crane-lib/pull/70/files 

DO NOT MERGE UNTIL GO.MOD is updated.

### DIFF
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -10,14 +10,11 @@ import (
 
 	"github.com/konveyor/crane-lib/state_transfer"
 	"github.com/konveyor/crane-lib/state_transfer/endpoint"
-	"github.com/konveyor/crane-lib/state_transfer/endpoint/route"
+	ls "github.com/konveyor/crane-lib/state_transfer/endpoint/local_service"
 	"github.com/konveyor/crane-lib/state_transfer/meta"
-	metadata "github.com/konveyor/crane-lib/state_transfer/meta"
 	"github.com/konveyor/crane-lib/state_transfer/transfer"
 	"github.com/konveyor/crane-lib/state_transfer/transfer/rsync"
-	"github.com/konveyor/crane-lib/state_transfer/transport"
-	"github.com/konveyor/crane-lib/state_transfer/transport/stunnel"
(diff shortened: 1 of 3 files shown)


cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -80,9 +87,14 @@ func NewTransferOptions(streams genericclioptions.IOStreams) *cobra.Command {
...
-	cmd.Flags().StringVar(&t.PVCNamespace, "pvc-namespace", "", "The namespace of the pvc which is to be transferred, if empty it will try to use the namespace in source-context, if both are empty it will error")
-	cmd.Flags().StringVar(&t.PVCName, "pvc-name", "", "The pvc name which is to be transferred on the source")
+	cmd.Flags().StringVar(&t.DestinationContext, "destination-context", "", "The name of destination context in current kubeconfig. Will use the default context")
+	cmd.Flags().StringVar(&t.SourcePVCNamespace, "source-pvc-namespace", "", "The namespace of the pvc which is to be transferred, if empty it will try to use the namespace in source-context, if both are empty it will error")
+	cmd.Flags().StringVar(&t.SourcePVCName, "source-pvc-name", "", "The pvc name which is to be transferred on the source")
+	cmd.Flags().StringVar(&t.DestinationPVCNamespace, "destination-pvc-namespace", "", "The namespace of the pvc which is to be transferred, if empty it will try to use the namespace in destination-context, if both are empty it will error")
+	cmd.Flags().StringVar(&t.DestinationPVCName, "destination-pvc-name", "", "The pvc name which is to be transferred on the source")
```
alaypatel07, line 94: can we change the help tool tip here?

```
"The pvc name which is to which the source data will be transferred to. If the name is empty, it will be the same as the source name"
```

cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -92,9 +104,20 @@ func (t *TransferPVCOptions) Complete(c *cobra.Command, args []string) error {
 		return err
 	}
```
alaypatel07, line 106: Can we add a block to default the destination PVC name to the source PVC name if empty?

PR #36, opened 2021-09-24
Add Github action for releasing crane binaries

shawn-hurley:
* creates three binaries linux/amd64, darwin/amd64, darwin/arm64 
* Uploads the binaries for the release
* allows anyone to trigger the action to create the release.
* TODO: Release notes creation for the release.


rayfordj (approved):
visack

sseago (approved):

PR #39, opened 2021-10-13
adding plugin-manager add, list and remove

JaydipGabani:
This PR includes changes that add the functionality of an additional `plugin-manager` subcommand under `crane`. [Here](https://github.com/konveyor/enhancements/tree/master/enhancements/crane-2.0/plugin-management) is the proposed enhancement for the same.

This PR is not the whole implementation of the proposed enhancement, but some. Following are the changes that are added.
- To be able to list the plugins from the hardcoded default repository
- To be able to add the plugin from the hardcoded default repository
- To be able to remove the installed plugin

Future work may be from the following PR or might be the modification of the current PR:
- To be able to retrieve metadata information of the plugin with list plugin command
- To be able to add the desired version of the plugin
- To be able to list the installed plugins
- To be able to override hardcoded default repository URL and override default plugin path
There might be more tasks related to validation and otherwise than what is mentioned above.

### DIFF
--- a/cmd/plugin-manager/add/add.go
+++ b/cmd/plugin-manager/add/add.go
@@ -1,0 +1,196 @@
+package add
+
+import (
+	"errors"
+	"fmt"
+	"github.com/konveyor/crane/internal/flags"
+	"github.com/konveyor/crane/internal/plugin"
+	"github.com/spf13/cobra"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+type Options struct {
+	globalFlags *flags.GlobalFlags
+	Repo        string
+	PluginDir   string
+	Version     string
+}
+
+func (o *Options) Complete(c *cobra.Command, args []string) error {
+	// TODO: @jgabani
+	return nil
+}
+
+func (o *Options) Validate(args []string) error {
+	// TODO: @jgabani
+	log := o.globalFlags.GetLogger()
+
+	pluginDir, err := filepath.Abs(fmt.Sprintf("%v/%v", o.PluginDir, o.Repo))
+	if err != nil {
+		return err
+	}
+
+	files, err := ioutil.ReadDir(pluginDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+
+	paths, err := plugin.LocateBinaryInPluginDir(o.PluginDir, args[0], files)
+	if err != nil {
+		return err
+	}
+
+	if len(paths) > 0 {
+		log.Errorf("The binary is already installed in the following path, either delete the binary or mention a repo from which the binary is needed")
+		for _, path := range paths {
+			fmt.Printf("%s \n", path)
+		}
+		return errors.New("the binary is already installed in the following path, either delete the binary or mention a repo from which the binary is needed")
+	}
+	return nil
+}
+
+func (o *Options) Run(args []string) error {
+	return o.run(args)
+}
+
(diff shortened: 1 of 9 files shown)


cmd/plugin-manager/list/list.go
```diff
@@ -0,0 +1,243 @@
...
+
+func (o *Options) run() error {
+	log := o.globalFlags.GetLogger()
+	// TODO: for multiple repo, read values from conf file to this map
+	repos := make(map[string]string)
+	repos[DEFAULT_REPO] = DEFAULT_REPO_URL
+
+	manifestMap := make(map[string]map[string]Manifest)
+
+	for repo, url := range repos {
```
shawn-hurley, line 113: I think I have seen this set of code twice, wonder if it should be a helper?

internal/plugin/plugin_helper.go
```diff
@@ -85,3 +85,7 @@ func isPluginInList(plugin transform.Plugin, list []string) bool {
 	}
 	return false
 }
+
+func IsExecAny(mode os.FileMode) bool {
```
shawn-hurley, line 89: This is interesting, if we are passing it through, why don't we make it public? 

@sseago

JaydipGabani: No need as of now, updated the plugin package for now and added helper functions there.

cmd/plugin-manager/list/list.go
```diff
@@ -0,0 +1,243 @@
...
+	"io/ioutil"
+	"net/http"
+	"os"
+	"runtime"
+	"text/tabwriter"
+)
+
+const (
+	DEFAULT_REPO     = "default"
+	DEFAULT_REPO_URL = "https://raw.githubusercontent.com/JaydipGabani/crane-plugins/main/index.yml"
```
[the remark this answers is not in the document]
JaydipGabani: Done

cmd/plugin-manager/list/list.go
```diff
@@ -0,0 +1,243 @@
...
+			}
+		}
+	}
+
+	printInformation(availablePlugin)
+}
+
+func printInformation(plugins map[string]AvailablePlugins) {
+	w := tabwriter.NewWriter(os.Stdout, 1, 1, 1, ' ', 0)
+	fmt.Fprintf(w, "Name \t ShortDescription \t AvailableVersions \n")
```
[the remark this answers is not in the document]
shawn-hurley: if that was the case I was thinking that you wouldn't have to add the tabs yourself?

cmd/plugin-manager/add/add.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+		}
+	} else {
+		// throw error saying that the plugin doest exists
+		log.Errorf(fmt.Sprintf("The plugin %s is not found", args[0]))
+	}
+	return nil
+}
+
+func DownloadBinary(filepath string, filename string, url string) error {
```
shawn-hurley, line 160: Should this be exported?

cmd/plugin-manager/add/add.go
```diff
@@ -0,0 +1,196 @@
...
+		return err
+	}
+	defer out.Close()
+
+	// Write the body to file
+	_, err = io.Copy(out, resp.Body)
+	return err
+}
+
+func BuildInstallBinaryPath(path string) string {
```
shawn-hurley, line 188: Should this be exported as well?

cmd/plugin-manager/add/add.go
JaydipGabani: this function is not used anywhere else, I can make it not exposed in the follow up PR.

PR #45, opened 2021-11-02
Fix pager list processing

alaypatel07:
depends on #42

### DIFF
--- a/cmd/export/discover.go
+++ b/cmd/export/discover.go
@@ -202,7 +202,7 @@ func getObjects(g *groupResource, namespace string, d dynamic.Interface, logger 
 	}
 
 	p := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
-		return c.Namespace(namespace).List(context.Background(), metav1.ListOptions{})
+		return c.Namespace(namespace).List(context.Background(), opts)
 	})
 
 	list, _, err := p.List(context.TODO(), metav1.ListOptions{})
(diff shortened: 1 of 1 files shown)


cmd/export/discover.go
```diff
@@ -202,17 +203,26 @@ func getObjects(g *groupResource, namespace string, d dynamic.Interface, logger
...
-		return nil, fmt.Errorf("expected unstructured.UnstructuredList type got %T for group: %s, kind: %s", l, g.APIGroup, g.APIResource.Kind)
+	unstructuredList := &unstructured.UnstructuredList{Items: []unstructured.Unstructured{}}
+	err = meta.EachListItem(list, func(object runtime.Object) error {
+		u, ok := object.(*unstructured.Unstructured)
+		if !ok {
+			// TODO: explore aggregating all the errors here instead of terminating the loop
+			logger.Errorf("expected unstructured.Unstructured but got %T for groupResource %s and object: %#v\n", g, object)
+			return fmt.Errorf("expected *unstructured.Unstructured but got %T", u)
+		}
+		unstructuredList.Items = append(unstructuredList.Items, *u)
```
shawn-hurley, line 221: Should we deep copy here or do we expect that not to matter because not a cache?

shawn-hurley (approved):
Just one question I don't think it will be that big of a deal

cmd/export/discover.go
alaypatel07: I think the 2nd. The only use-cases I have in mind right now are read-only so no need to pay the price for a copy?

shawn-hurley: +1 makes sense

PR #46, opened 2021-11-04
changing default plugin dir

JaydipGabani:


cmd/plugin-manager/add/add.go
```diff
@@ -46,7 +46,7 @@ func (o *Options) Validate(args []string) error {
 	if len(args) != 1 {
 		return errors.New("please input only one plugin name")
 	}
-
+	o.PluginDir = fmt.Sprintf("%v/%v", o.PluginDir, plugin.MANAGED_DIR)
```
sseago, line 49: This seems risky. If  Validate ever got called twice, o.PluginDir might end up being "plugins/managed/managed" -- I think o.PluginDir should probably stay as-is but maybe define a func:
```
func (o *Options) ManagedPluginDir() string {
    return fmt.Sprintf("%v/%v", o.PluginDir, plugin.MANAGED_DIR)
}
```

sseago: Then use `o.ManagedPluginDir()` instead of `o.PluginDir` any time you need the actual directory where managed plugins go.

cmd/plugin-manager/list/list.go
```diff
@@ -96,14 +96,20 @@ func addFlagsForOptions(o *Flags, cmd *cobra.Command) {
...
+	o.PluginDir = fmt.Sprintf("%v/%v", o.PluginDir, plugin.MANAGED_DIR)
 	if o.Installed {
 		// retrieve list of all the plugins that are installed within plugin dir
+		// TODO: differentiate between multiple repos
 		plugins, err := plugin.GetFilteredPlugins(o.PluginDir, []string{}, log)
 		if err != nil {
 			return err
 		}
-		fmt.Println(fmt.Sprintf("Listing plugins from path - %s", o.PluginDir))
+		if len(plugins) > 1 {
```
sseago, line 107: I think this might be better with the original. Otherwise if we add/remove default plugins, this language changes. How about just "Listing configured plugins from path", that way "configured" can mean the plugins found here and any defaults not needed to be installed.

sseago (changes requested):

sseago (approved):

PR #48, opened 2021-11-06
lowercase optional arg names before passing them to runner.

sseago:
The viper config file parser lower-cases map keys. To be consistent between
optionals via config file and optionals on the command-line, we need to
downcase the keys passed in to the plugin runner. As a result, when plugins
retrieve extras by key, they will need to downcase the string they're checking for.


JaydipGabani (approved):
LGTM

djzager (approved):

PR #49, opened 2021-11-10
changing default source repository branch

JaydipGabani:


sseago (commented):
So I think this is fine, but we should make sure that "release" is the right name for the branch. It's not tied to a particular release, but it is the branch that upstream crane will point to to download plugins. I wonder whether "release" might be a confusing name. Any ideas for a better one, though?

sseago (approved):

PR #50, opened 2021-11-10
updating to crane-lib 0.0.3

sseago:


sseago:
I've gone through my sprint demo script with the 0.0.3 dep and it all works.

djzager (approved):

PR #51, opened 2021-11-12
Making sure that plugin-manager works in disconnected env with local crane-plugins

JaydipGabani:


internal/plugin/plugin_manager_helper.go
```diff
@@ -86,51 +87,81 @@ func BuildManifestMap(log *logrus.Logger, name string, repoName string) (map[str
...
 
 // takes url as input and fetches the manifest of a plugin
-func YamlToManifest(url string) (Manifest, error) {
+func YamlToManifest(URL string) (Manifest, error) {
 	plugin := Manifest{}
-	res, err := http.Get(url)
-	if err != nil {
-		return plugin, err
-	}
+	if !IsUrl(URL) {
```
djzager, line 127: I personally think that instead of having `IsUrl` in each of these functions for handling 1) the index, 2) the manifest, 3) the binary that you could instead create a new function `GetBytes(URL string) ([]byte, error)` for example that just takes the URL (whether that be `file:///some/path`, `https://raw.github...`, `/some/path`) and returns the bytes.

djzager (changes requested):
There are a few changes that I would like to see in this PR:

1. Instead of `IsUrl` and making the decision each time whether to `ReadFile` or get the URL and then `ReadAll`, have one function that just returns the bytes that we can unmarshal. This especially applies in the case of `YamlToManifest` and `GetYamlFromUrl`.
2. Make sure that we can handle `file://` urls
3. Add unit tests so that we clearly define the plugin-managers behavior on this topic...prevent us from having regressions in the future.

sseago (changes requested):
I think these are the right changes to the functionality. We still need unit tests to verify that https, file, and raw path entries work with the new code.

sseago (approved):

djzager (commented):
VISACK

djzager (approved):

PR #63, opened 2021-12-09
hack: provide kube version override for minikube

djzager:
In one particular scenario, I need an old kube and need to be able to override it.


alaypatel07 (approved):
ACK

PR #69, opened 2021-12-22
feat: lay down kustomization in apply

djzager:
Include kustomization.yaml in the output directory so that it can be
used with kustomize directly (overriding namespace etc.).


djzager:
cc @eriknelson @shawn-hurley @sseago

This is the idea I would like to build on when we talk about leveraging kustomize in crane. I'm going to leave this as a draft as it needs to be introduced first as an enhancement.

djzager:
This implements the enhancement described in https://github.com/konveyor/enhancements/pull/57

djzager:
Closing as the related enhancement was rejected.

PR #75, opened 2022-03-17
Add missing pieces required for stateful migrations

pranavgaikwad:
- Adds code to discover nodes on which source PVC is mounted
- Adds code to parse PVC name & namespace mappings:
  - This is based on David's alpha work where mappings are passed as `:` separated strings instead of separated flags
- Adds new flag to take Storage Class of destination PVC as input
- Adds new flag to take capacity of destination PVC as input
- Adds new flag to enable checksum verification for transfer


cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -45,24 +46,77 @@ type TransferPVCOptions struct {
...
-	DestinationContext string
-	PVCName            string
-	PVCNamespace       string
-	Endpoint           string
+	logger               logrus.FieldLogger
+	SourceContext        string
+	DestinationContext   string
+	PVCName              mappedName
+	PVCNamespace         mappedName
+	DestStorageClassName string
```
alaypatel07, line 54: I wonder if we should convert these flags into an internal data structure like:
```
apiVersion: config.transfer-pvc.crane.dev/v1alpha1
kind: PVCConfig
spec:
    source: <ObjectReference>
    claimTemplate: <PVCTemplate>
    verify: <bool>
    ...
```

1. read the config file and decode the above values
2. if flags provided override the flag values specs
3. Then use the above object to actually transfer the PVC. 

IMO it will abstract out the input with transfer action nicely. Not suggesting doing all in this PR, open to how to break it up if this sounds like a good idea

cc @djzager not sure if you had opinions, just wanted to loop you in.

djzager: I think, at a minimum, we can put these into a struct that represents what a `PVCConfigSpec` would be. Since I believe concrete API types like that are probably more appropriate in crane-lib than in the cli directly.

I would support us creating an issue to convert such a struct to a more clearly defined API type in crane-lib. My recommendation for this PR would be to create a new `PVCConfigSpec` with this stuff.

cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -89,10 +143,13 @@ func NewTransferOptions(streams genericclioptions.IOStreams) *cobra.Command {
 
 func addFlagsForTransferPVCOptions(t *TransferPVCOptions, cmd *cobra.Command) {
 	cmd.Flags().StringVar(&t.SourceContext, "source-context", "", "The name of the source context in current kubeconfig")
-	cmd.Flags().StringVar(&t.DestinationContext, "destination-context", "", "The name of destination context current kubeconfig")
-	cmd.Flags().StringVar(&t.PVCNamespace, "pvc-namespace", "", "The namespace of the pvc which is to be transferred, if empty it will try to use the namespace in source-context, if both are empty it will error")
-	cmd.Flags().StringVar(&t.PVCName, "pvc-name", "", "The pvc name which is to be transferred on the source")
+	cmd.Flags().StringVar(&t.DestinationContext, "destination-context", "", "The name of destination context in current kubeconfig")
+	cmd.Flags().Var(&t.PVCNamespace, "pvc-namespace", "The namespace of the pvc which is to be transferred, if empty it will try to use the namespace in source-context, if both are empty it will error")
+	cmd.Flags().Var(&t.PVCName, "pvc-name", "The pvc name which is to be transferred on the source")
```
djzager, line 148: I don't see anything specifically calling out the `src:dest` nature of these parameters. We should include that.

cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -193,16 +256,15 @@ func (t *TransferPVCOptions) run() error {
 
 	// set up the PVC on destination to receive the data
 	pvc := &corev1.PersistentVolumeClaim{}
-	err = srcClient.Get(context.TODO(), client.ObjectKey{Namespace: t.PVCNamespace, Name: t.PVCName}, pvc)
+	err = srcClient.Get(context.TODO(),
+		client.ObjectKey{
+			Namespace: t.PVCNamespace.source,
+			Name:      t.PVCName.source}, pvc)
```
djzager, line 262:
```suggestion
	err = srcClient.Get(
		context.TODO(),
		client.ObjectKey{
			Namespace: t.PVCNamespace.source,
			Name:      t.PVCName.source
		},
		pvc,
	)
```

nit

cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -502,6 +598,18 @@ func createAndWaitForRoute(pvc *corev1.PersistentVolumeClaim, destClient client.
 	return e
 }
 
+// getDestinationPVC given a source PVC, returns a PVC to be created in the destination cluster
+func (t *TransferPVCOptions) getDestinationPVC(sourcePVC *corev1.PersistentVolumeClaim) *corev1.PersistentVolumeClaim {
+	pvc := sourcePVC.DeepCopy()
+	clearDestPVC(pvc)
```
djzager, line 604: looks like this is only used here in `getDestinationPVC`, can we merge what's in that function with what you have below?

djzager (commented):
Just a few comments. Like it a lot.

cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -44,23 +45,41 @@ const (
...
+	name mappedNameVar
+	// namespace defines namespace of the PVC,
+	// mapped in format <source>:<destination>
+	namespace mappedNameVar
+	// storageClassName defines storage class of destination PVC
+	storageClassName string
+	// storageRequests defines requested capacity of destination PVC
+	storageRequests quantityVar
+}
+
```
djzager, line 74: :+1:

djzager (approved):

PR #76, opened 2022-03-17
bug: skip 'failures' directory when reading files

djzager:
This makes it so that we don't attempt to process `failures`.


jmontleon (approved):
ACK!

PR #80, opened 2022-04-16
Corrected the typo  `crane export --help`

apoorvajagtap:
Correct the typo "example" in the description of `oc export --as-extras`. This meets the 1st point in https://github.com/konveyor/crane/issues/71

I have just started exploring crane, so in case the 2 & 3 points in https://github.com/konveyor/crane/issues/71 are also required, I'd be happy to add those as well in the same PR. Let me know, thanks!


pranavgaikwad (approved):
Thanks for the fix

pranavgaikwad:
@apoorvajagtap Let's fix 2 and 3 in the issue in this PR.

PR #90, opened 2022-05-13
bug: fix optional flags handling

djzager:


sseago:
Has this been tested with file input?  Changing the map to a string array may break yaml input files specified as maps. Also, this used to work with map input. What changed? What's driving the change here?

PR #92, opened 2022-05-16
update cli doc to make it clear how to use slice/map params

sseago:
As we discussed earlier on slack, we may still want to revisit this to see if we can find a way to refactor this to remove the need for the extra layer of quoting, but in the meantime, this doc update makes it clear how to input doc and map params in the optional-flags list.


cmd/transform/transform.go
```diff
@@ -150,6 +150,9 @@ func (o *Options) run() error {
 	}
 	if len(o.OptionalFlags) > 0 {
 		runner.OptionalFlags = optionalFlagsToLower(o.OptionalFlags)
+		for k, v := range runner.OptionalFlags {
+			log.Infof("optionalarg k: %v, v: %v", k, v)
+		}
```
djzager, line 155: I think this could simply be:
```suggestion
		lob.Debugf("parsed optional-fags: %v", runner.OptionalFlags)
```

was playing around with it and this is what it would look like:
```
$ ./crane transform --export-dir=export --transform-dir=transform --optional-flags="add-annotations=migration-using-crane=true,foo=bar,baz=zoo" --debug
DEBU[0000] optional-flags: map[add-annotations:migration-using-crane=true baz:zoo foo:bar]
```

djzager (changes requested):
1 nit

cmd/transform/transform.go
sseago: Actually, that was debug code that I had intended to remove. But I could keep it as `Debugf` as you suggest. Will update this with the change.

sseago: updated

djzager (approved):
/lgtm

PR #114, opened 2022-06-16
Use rsync options struct to override the default rsync options in library

### LINKED ISSUE #105: [BUG] Update Rsync transfer in pvc-transfer to use custom options when upstream PR is merged
After integration with pvc-transfer, custom rsync options were removed from pvc-transfer subcommand, we need to add them back when upstream PR is merged https://github.com/backube/pvc-transfer/pull/19

pranavgaikwad:
Fixes #105 

- Drops ALL capabilities from rsync pods, this allows running the Pod with PS admission enforcement
- Drops Rsync options that require privileged containers, containers will now run as restricted by default
  - This is based on OpenShift's SCC capability to assign Rsync pods right UID/GID on the destination cluster
- Created a follow up issue to allow users set custom UID/GIDs and optionally run privileged pods


cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -413,11 +431,32 @@ func (t *TransferPVCCommand) run() error {
...
+				verbose(true),
+			),
+			ContainerSecurityContext: corev1.SecurityContext{
+				Privileged: &falseBool,
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{"ALL"},
+				},
+				RunAsNonRoot: &trueBool,
+				RunAsUser:    clientPodSecCtx.RunAsUser,
+			},
```
JaydipGabani, line 456:
```suggestion
                                AllowPrivilegeEscalation: &falseBool,
				SeccompProfile: &corev1.SeccompProfile{
								Type: "RuntimeDefault",
							},
			},
```

I think we need these two field values on container/pod SC as well.

cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -389,11 +390,28 @@ func (t *TransferPVCCommand) run() error {
...
 		context.TODO(),
 		destClient,
 		logger, destPVCList, stunnelServer, e, labels, nil, rsyncPassword,
-		transfer.PodOptions{},
+		transfer.PodOptions{
+			ContainerSecurityContext: corev1.SecurityContext{
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{"ALL"},
+				},
+				RunAsNonRoot: &trueBool,
```
JaydipGabani, line 409: Might need to add the below-mentioned fields here as well.

cmd/transfer-pvc/transfer-pvc.go
pranavgaikwad: @JaydipGabani Thanks! didn't know about these two

djzager: Does this stuff exist in prior clusters? Do we need to be careful about what we apply to containers on the Source side for fear of them not coming up?

@jmontleon mentioned something to this effect recently and I just wonder if it applies here.

JaydipGabani: I saw pods getting admitted in the 3.11 cluster with this `SecurityContext`, `SeccompProfile` field gets omitted as it doesn't exist.

djzager: Cool. That's great news if we don't have to worry about mismatch between src/dest clusters for transfer-pvc.

JaydipGabani (approved):
/lgtm

PR #125, opened 2022-06-30
Crane sub-command to run function

MundraAnkur:


cmd/run-function/runfn/runfn.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package runfn
+//Reference: sigs.k8s.io/kustomize/kyaml/runfn
```
djzager, line 5: I may be missing it. But this looks like something we could just include and use as a dependency.

https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.13.7/runfn

cmd/run-function/runfn/runfn.go
```diff
@@ -0,0 +1,138 @@
...
+	// Output can be set to write the result to Output
+	Output io.Writer
+
+	// Env contains environment variables that will be exported to container
+	Env []string
+
+	// functionFilterProvider provides a filter to perform the function.
+	// this is a variable, so it can be mocked in tests
+	functionFilterProvider func(filter runtimeutil.FunctionSpec, fnConfig *yaml.RNode) (kio.Filter, error)
+}
```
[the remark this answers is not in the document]
djzager: What I meant was, does something break if you `import` their `runfn` package without using all of the `RunFn` struct members?

My personal preference would be to import wherever possible, if we must fork in order to support `podman` then we should do that and look to contribute upstream.

cmd/run-function/runfn/runfn.go
djzager: Like I said in [the comment below](https://github.com/konveyor/crane/pull/125#discussion_r912159637), my personal preference would be to import modules wherever possible. If you were to create a fork, then a couple of things:
1. If they make changes we can bring those into the fork with `rebase`.
2. We can submit the work needed to support `podman` "upstream", be good community citizens, and all that jazz.

cmd/run-function/runfn/runfn.go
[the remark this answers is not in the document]
MundraAnkur: I've not tried using their RunFns, but I don't think it will break anything, we need to adjust how we pass in the required configuration. I'll try using import






> My personal preference would be to import wherever possible, if we must fork in order to support `podman` then we should do that and look to contribute upstream.

Do you mean adding a PR to kustomize?

.github/workflows/go117.yml
```diff
@@ -0,0 +1,27 @@
+name: Build and test Go with 1.17
```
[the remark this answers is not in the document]
MundraAnkur: @djzager Yeah, I'll remove that when we move from draft to PR, this was just to test if everything works with Go1.17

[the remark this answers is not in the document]
MundraAnkur: We can actually, update the already available action in `go.yml`

cmd/run-function/runfn/runfn.go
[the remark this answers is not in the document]
djzager: > Do you mean adding a PR to kustomize?

yup

cmd/run-function/util.go
```diff
@@ -0,0 +1,89 @@
...
+// https://github.com/docker/distribution/blob/master/reference/reference.go. A valid
+// name definition is:
+//	name                            := [domain '/'] path-component ['/' path-component]*
+//	domain                          := domain-component ['.' domain-component]* [':' port-number]
+//	domain-component                := /([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])/
+//	port-number                     := /[0-9]+/
+//	path-component                  := alpha-numeric [separator alpha-numeric]*
+// 	alpha-numeric                   := /[a-z0-9]+/
+//	separator                       := /[_.]|__|[-]*/
+//https://github.com/GoogleContainerTools/kpt/blob/main/pkg/api/kptfile/v1/validation.go#L120-L150
```
djzager, line 57: nit. use a permalink

MundraAnkur: done!

MundraAnkur: Okay, I didn't know this but thanks for adding it here.
Now, I've updated it correctly

cmd/run-function/runcmd.go
```diff
@@ -0,0 +1,217 @@
+package run_function
```
djzager, line 1: Annoying but important nitpick. I think the package should be `runfn` like the command, and the file should be `runfn.go`.

djzager (commented):
I really like what I see. This is good. I am curious now how much these conversations about binary plugins impacts what is being done.

cmd/run-function/runcmd.go
MundraAnkur: done!

MundraAnkur:
> I really like what I see. This is good. I am curious now how much these conversations about binary plugins impact what is being done.

I think it won't affect the existing work, but yes when we will add support for binaries, almost all functions will be updated to accommodate binary execution

djzager (approved):

PR #136, opened 2022-07-15
crane sub-command to run krm functions

MundraAnkur:


cmd/runfn/runfn.go
```diff
@@ -0,0 +1,217 @@
...
+	TransformedContent bytes.Buffer
+}
+
+func NewFnRunCommand(f *flags.GlobalFlags) *cobra.Command {
+	o := &Options{
+		cobraGlobalFlags: f,
+	}
+	cmd := &cobra.Command{
+		Use:          "runfn [IMAGE] [flags] [--args]",
+		Short:        "Transform resources by executing function \nExperimental: This command is under active development and may change without notice.",
```
djzager, line 39:
```suggestion
		Short:        "Transform resources by executing KRM function \n\nExperimental: This command is under active development and may change without notice.",
```

djzager (changes requested):
One change, otherwise lgtm

cmd/runfn/runfn.go
MundraAnkur: 1. Done!
2. Added `Long`  command message, as the short message appears in `crane -v` and `crane -h`

djzager (approved):

Issue #148, opened 2023-02-28
[RFE] Add option to export cluster-scoped RBAC resources


dmartinol:
**What is your persona?**

I'm an application developer that is designing a POC of an open-source application that integrates crane to export the configuration of an application.

**What is your story / use case?**

As a user of the `crane export` command, I want to include all the non-namespaced (e.g. cluster scoped) RBAC resources that are linked to the `ServiceAccount` defined in the exported namespace(s), so that the generated manifests can be deployed on a new environment with no further manual operations.

Using an optional flag to the `crane export` command, like `--cluster-scoped-rbac` (defaulted to "false" to disable the requested feature), at the end of the crane export-transform-apply pipeline, we can generate a complete and consistent set of manifests that can be deployed on any environment without any manual intervention to setup the RBAC aspects.

**Do you have a suggestion for implementation?**

I would implement this as an optional flag in the crane core module, since the export command does not support the concept of plugin. 

Since we want only some of these cluster-scoped resources, we should define somewhere the list of managed types (e.g., `ClusterRole`, `ClusterRoleBinding` and, in case of OpenShift cluster, also `SecurityContextConstraints`) and load them only after all the configured namespaces have been exported, to discard those that are not connected to any of the exported `ServiceAccount` (maybe a cache can be required to store all of them, or at least their namespace and name pairs).

Once selected, those manifests can be stored in a dedicated folder, with a name that can't match a namespace name, like `_cluster`.

While the contributors review the issue, I would start developing the feature to validate the proposed solution.

PR #149, opened 2023-03-03
Issue #148: introduce cluster-scoped-rbac

dmartinol:
Added `ClusterScopeHandler` struct to manage cluster scoped resources (initially with `filterRbacResources` function but later can be extended with other functions e.g. to export also CRD configurations).
This is triggered by the optional flag `cluster-scoped-rbac` in the `crane export` command.

Notes: 
* Updated Go version to 1.18 due to the `golang.org/x/exp/slices` package.
* Since a new flag has been added to the `export` command, the documentation should also reflect the same. Please let me know how to do update the documentation as well


shawn-hurley (changes requested):
I wonder if we can move this out of cmd package and into a pkg for re-use. 

In the past, we would have put most of this in crane-lib but if this is going to work for your usecase, I don't have a huge problem.

The things that I would for sure change is that when listing ClusterScoped resources, we should only list the ones that we care about not all of them coming from discovery, I think that will put some strain on API servers if this is in an operator like git-ops primer.

cmd/export/cluster.go
```diff
@@ -0,0 +1,221 @@
...
+	clusterScopeHandler := &ClusterScopeHandler{}
+
+	return clusterScopeHandler
+}
+
+func (c *ClusterScopeHandler) filterRbacResources(resources []*groupResource, log logrus.FieldLogger) []*groupResource {
+	log.Debug("Looking for ServiceAccount resources")
+
+	handler := NewClusterScopedRbacHandler(log)
+	for i := len(resources) - 1; i >= 0; i-- {
```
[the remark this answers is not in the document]
dmartinol: The reason for using reverse loop is that we remove the resources of the admitted kinds (L41) which alters the content of the initial array: the reverse navigation avoids to skip some of the elements because the elements with index from 0-i always remain unchanged.
to use range I think we need to create a temporary slice and populate it by copying the elements. if you agree, I can do that way (should be a copy of references,  so not a big issue in terms of memory)

cmd/export/discover.go
```diff
@@ -201,18 +206,25 @@ func resourceToExtract(namespace string, labelSelector string, dynamicClient dyn
...
-	if !g.APIResource.Namespaced {
-		return &unstructured.UnstructuredList{}, nil
-	}
-
 	p := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
-		return c.Namespace(namespace).List(context.Background(), opts)
+		if g.APIResource.Namespaced {
+			return c.Namespace(namespace).List(context.Background(), opts)
+		} else {
+			return c.List(context.Background(), opts)
```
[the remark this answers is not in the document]
dmartinol: You mean adding ListOptions so that, for instance, when we collect the ServiceAccounts we also collect the meaningful ClusterRoleBinding and ClusterRole by using FieldSelector queries to fetch only the relevant instances?
If this is the purpose of the comment, it would require more changes to the discovery module because ATM if fetches only resources of a given, single kind. If we return resources of multiple kinds, instead, the code using getObjects function would need more changes.
Please let me know what was your actual goal.

cmd/export/cluster.go
```diff
@@ -0,0 +1,221 @@
...
+		FromUnstructured(clusterResource.Object, &scc)
+	if err != nil {
+		c.log.Warnf("Cannot convert to securityv1.SecurityContextConstraints: %s", err)
+	} else {
+		for _, f := range c.filteredClusterRoleBindings.objects.Items {
+			var crb authv1.ClusterRoleBinding
+			err := runtime.DefaultUnstructuredConverter.
+				FromUnstructured(f.Object, &crb)
+			if err != nil {
+				c.log.Warnf("Cannot convert to authv1.ClusterRoleBinding: %s", err)
```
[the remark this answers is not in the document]
dmartinol: Resolved in latest commit

cmd/export/cluster.go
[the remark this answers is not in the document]
dmartinol: Resolved in latest commit

cmd/export/discover.go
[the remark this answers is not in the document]
dmartinol: FYI, not all the fields are eligible to be used in the ListOptions.FieldSelector, e.g. the following command:
oc get clusterrolebindings.rbac.authorization.k8s.io --field-selector roleRef.name=AAA
would actually return an error because of that:
Error from server (BadRequest): Unable to find "rbac.authorization.k8s.io/v1, Resource=clusterrolebindings" that match label selector "", field selector "roleRef.name=AAA": "roleRef.name" is not a known field selector: only "metadata.name", "metadata.namespace"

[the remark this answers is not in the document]
shawn-hurley: I think what I was asking, is that we look at the kind, and only retrieve the kinds of RBAC and SCC rather than filter here.

cmd/export/cluster.go
```diff
@@ -0,0 +1,243 @@
...
+			if c.acceptClusterRoleBinding(crb) {
+				filteredClusterRoleBindings.Items = append(filteredClusterRoleBindings.Items, crb)
+				c.log.Infof("Found matching %s %s", crb.GetKind(), crb.GetName())
+			}
+		}
+		c.filteredClusterRoleBindings.objects = &filteredClusterRoleBindings
+	}
+}
+
+func (c *ClusterScopedRbacHandler) filteredResourcesOfKind(resource admittedResource) (*groupResource, bool) {
```
djzager, line 111: This filter logic, to me, seems more appropriate for the `transform` step.

cmd/export/cluster.go
```diff
@@ -0,0 +1,243 @@
...
+			continue
+		}
+
+		if crb.RoleRef.Kind == "SecurityContextConstraints" && crb.RoleRef.Name == scc.Name {
+			c.log.Infof("Accepted %s of kind %s", clusterResource.GetName(), clusterResource.GetKind())
+			return true
+		} else {
+			sccSystemName := fmt.Sprintf("system:openshift:scc:%s", clusterResource.GetName())
+			if crb.RoleRef.Kind == "ClusterRole" && crb.RoleRef.Name == sccSystemName {
+				c.log.Infof("Accepted %s of kind %s (match wia ClusterRoleBinding %s)",
```
djzager, line 212:
```suggestion
				c.log.Infof("Accepted %s of kind %s (match via ClusterRoleBinding %s)",
```

djzager (changes requested):
This is awesome work and thank you for putting it together. I do wonder though if it may be better for us to have the `cluster-scoped-rbac` flag (as you have included here) and when it is set to `true`, we go ahead and "GET" the `admittedClusterScopeResources`, write these resources out in the `_cluster` dir you create, leaving the filter logic you have in `discover.go` for the transform step.

My main concern here is that the filtering of RBAC resources looks a lot like a `transform` and I think it's wise for us to keep `export` and `transform` as distinct steps.

The part I have to apologize for is that I believe that means, in addition to changes here in `crane`, you'll need changes to:
1. The kubernetes plugin for ClusterRole(Binding)s https://github.com/migtools/crane-lib/blob/main/transform/kubernetes/kubernetes.go
1. The openshift plugin to handle security context constraints https://github.com/migtools/crane-plugin-openshift/blob/main/cmd.go

What do you think? I suspect the hardest thing will be related to the benefit you have here of being able to traverse all the resources and then pick the ones you want based on filter conditions. Might get tricky.

cmd/export/cluster.go
dmartinol: Hey @djzager, I perfectly agree with you that this seems to be a `transform` step, and you almost anticipated some of the reasons why I haven't followed this path (e.g., complexity and more repos impacted), but at the moment this can't be done without adding any context related to the other exported resources, because the PluginRun interface works only on individual resources, but the decision whether to accept or whiteout an RBAC resource depends also on other resources (e.g., `is there any ServiceAccount matching this ClusterRoleBinding`?). In brief, we should pass some kind of "context" to the plugin execution which, in turn, for each and every RBAC resource, should unmarshal all the files in all the exported namespaces to look for these matches.
I'm totally worried by performance, as you can imagine, this is why the initial implementation, even if it broke the usual 3-steps approach, looked more promising.
WDYT?

cmd/export/cluster.go
```diff
@@ -0,0 +1,244 @@
...
+	{Kind: "SecurityContextConstraints", APIgroup: "security.openshift.io"},
+}
+
+func NewClusterScopeHandler() *ClusterScopeHandler {
+	clusterScopeHandler := &ClusterScopeHandler{}
+
+	return clusterScopeHandler
+}
+
+func isClusterScopedResource(apiGroup string, kind string) bool {
```
djzager, line 34:
```suggestion
func isAdmittedClusterScopedResource(apiGroup string, kind string) bool {
```

cmd/export/cluster.go
djzager: Fair point. I believe the right answer for us in this scenario would be to articulate why we can't put this in transform plugin(s) via GitHub issue and proceed with this PR as it is currently implemented.

cmd/export/cluster.go
djzager: Maybe `isAdmitted` is even better, currently there is a mismatch between the name of the function and what it's doing.

djzager (commented):
I believe I follow what's going on and why. Hope to give this a final read through before COB today.

cmd/export/cluster.go
shawn-hurley: I am plus one, a follow up issue would be great

shawn-hurley (approved):
I am good to go, @djzager want to merge after your read through?

PR #151, opened 2023-03-07
150 segmentation violation

dmartinol:
Added initialization of filteredClusterRoleBindings in case the cluster resources are not accessible


cmd/export/cluster.go
```diff
@@ -106,6 +107,16 @@ func (c *ClusterScopedRbacHandler) prepareForFiltering() {
 			}
 		}
 		c.filteredClusterRoleBindings.objects = &filteredClusterRoleBindings
+	} else {
```
shawn-hurley, line 110: I would remove the else block and and a return statement below 109

Otherwise, this makes sense to me

shawn-hurley (approved):
Sorry for the late look just got through emails

Let me know if you need something else :)

cmd/export/cluster.go
dmartinol: fixed in next commit

djzager (approved):

PR #179, opened 2026-03-09
Fix nil pointer dereference on TransferredData in progress status check

RanWurmbrand:
## Summary

- Add nil check for `TransferredData` in `Progress.Status()` to prevent panic when rsync fails without transferring any data

## Problem

When `transfer-pvc` rsync retries are exhausted, the progress tracking code accesses `p.TransferredData.val` without checking if `TransferredData` is nil. This causes a nil pointer dereference panic instead of a clean "Failed" status.

## Fix

Added a nil check for `p.TransferredData` before accessing `.val` in `progress.go:304`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a potential crash in PVC transfer progress tracking when transfer data is unavailable.

### DIFF
--- a/cmd/transfer-pvc/progress.go
+++ b/cmd/transfer-pvc/progress.go
@@ -301,7 +301,7 @@ func (p *Progress) Status() status {
 			return succeeded
 		}
 		if p.TransferredFiles == 0 &&
-			p.TransferredData.val == 0 &&
+			(p.TransferredData == nil || p.TransferredData.val == 0) &&
 			p.TotalFiles == nil {
 			return failed
 		}


cmd/transfer-pvc/progress.go
```diff
@@ -301,7 +301,7 @@ func (p *Progress) Status() status {
 			return succeeded
 		}
 		if p.TransferredFiles == 0 &&
-			p.TransferredData.val == 0 &&
+			(p.TransferredData == nil || p.TransferredData.val == 0) &&
```
aufi: If it is at least little possible, we need avoid null/nil pointer for `p.TRansferredData` instead of workaround when it already happen.

RanWurmbrand: sounds good, should i close this pr for now, and we will back to that when we start work on the transfer-pvc command, or should i find a better solution, what do you think?
and sorry for super late replay, totally missed you'r comment

PR #194, opened 2026-03-27
Automate MTC-153 : empty PVC migration 

### LINKED ISSUE #273: Automate MTC-153 : empty PVC migration

nachandr:
Resolves [Issue](https://github.com/migtools/crane/issues/273)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end test coverage for empty PVC migration using non-admin credentials: verifies detection of empty PVCs, runs the export/transform/apply pipeline as a non-admin, transfers PVCs to the target cluster, applies generated manifests, waits for target app validation, and confirms migrated app data remains empty.

### DIFF
--- a/e2e-tests/tests/empty_pvc_migration_test.go
+++ b/e2e-tests/tests/empty_pvc_migration_test.go
@@ -3,6 +3,9 @@ package e2e
 import (
 	"fmt"
 	"log"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/konveyor/crane/e2e-tests/config"
 	. "github.com/konveyor/crane/e2e-tests/framework"
@@ -12,7 +15,7 @@ import (
 
 var _ = Describe("Empty PVC migration", func() {
 	It("[MTC-153] Migrate an empty PVC associated with an application", Label("tier0"), func() {
-		appName := "volumesnapshot"
+		appName := "app-with-empty-pvc"
 		namespace := appName
 		scenario := NewMigrationScenario(
 			appName,
@@ -22,22 +25,49 @@ var _ = Describe("Empty PVC migration", func() {
 			config.SourceContext,
 			config.TargetContext,
 		)
-		srcApp := scenario.SrcApp
-		tgtApp := scenario.TgtApp
-		kubectlSrc := scenario.KubectlSrc
-		kubectlTgt := scenario.KubectlTgt
+		if scenario.KubectlSrcNonAdmin.Context == "" {
+			Skip("source-nonadmin-context is required for non-admin test")
+		}
+		if scenario.KubectlTgtNonAdmin.Context == "" {
+			Skip("target-nonadmin-context is required for non-admin test")
+		}
+		srcApp := scenario.SrcAppNonAdmin
+		tgtApp := scenario.TgtAppNonAdmin
+
+		srcApp.ExtraVars = map[string]string{
+			"non_admin_user": "true",
+		}
+		tgtApp.ExtraVars = map[string]string{
+			"non_admin_user": "true",
+		}
+
+		By("Grant ns admin permissions to nonadmin user on source and target")
+		kubectlSrcNonAdmin, kubectlTgtNonAdmin, cleanup, err := SetupNamespaceAdminUsersForScenario(scenario, namespace)
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(cleanup)
 
 		By("Prepare source app")
 		log.Printf("Preparing source app %s in namespace %s\n", srcApp.Name, srcApp.Namespace)
-		Expect(PrepareSourceApp(srcApp, kubectlSrc)).NotTo(HaveOccurred())
+		Expect(PrepareSourceApp(srcApp, kubectlSrcNonAdmin)).NotTo(HaveOccurred())
 		log.Printf("Source app %s prepared successfully\n", srcApp.Name)
 
 		paths, err := NewScenarioPaths("crane-export-*")
 		Expect(err).NotTo(HaveOccurred())
 		DeferCleanup(func() {
-			By("Cleanup source and target resources")
-			if err := CleanupScenario(paths.TempDir, srcApp, tgtApp); err != nil {
-				log.Printf("cleanup: %v", err)
+			By("Cleanup temp directory")
+			if paths.TempDir != "" {
+				log.Printf("Removing temp dir: %s\n", paths.TempDir)
+				if err := os.RemoveAll(paths.TempDir); err != nil {
+					log.Printf("cleanup: failed to remove temp dir %q: %v", paths.TempDir, err)
+				}
+			}
+		})
+		DeferCleanup(func() {
+			By("Delete test namespace on source and target (best effort)")
+			for _, k := range []KubectlRunner{scenario.KubectlSrc, scenario.KubectlTgt} {
+				if _, err := k.Run("delete", "namespace", namespace, "--ignore-not-found=true", "--wait=false"); err != nil {
+					log.Printf("cleanup: failed to delete namespace %q on context %q: %v", namespace, k.Context, err)
+				}
 			}
 		})
 		By("List pvcs in the source namespace")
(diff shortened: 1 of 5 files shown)


e2e-tests/tests/empty_pvc_migration_test.go
```diff
@@ -0,0 +1,188 @@
...
+		err = filepath.Walk(paths.OutputDir, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.IsDir() || !strings.HasSuffix(path, ".yaml") {
+				return nil
+			}
+			filename := filepath.Base(path)
+
+			// Remove OpenShift auto-injected resources and RBAC bindings
```
aufi: I think this needs to be solved within crane itself (using OpenShiftPlugin if that supports it), please look if openshift plugin that could solve it already (https://github.com/migtools/crane-plugin-openshift) or open an issue to update it.

aufi: Possible OpenShift plugin update should get covered by https://github.com/migtools/crane/issues/264, but not stopping you if you'd want update it for this test.

e2e-tests/tests/empty_pvc_migration_test.go
```diff
@@ -0,0 +1,188 @@
...
+		pvcs, err := ListPVCs(srcApp.Namespace, "", srcApp.Context)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(pvcs).NotTo(BeEmpty(), "expected at least one pvc in source namespace %q", srcApp.Namespace)
+		log.Printf("Found %d pvcs in namespace %q", len(pvcs), srcApp.Namespace)
+		for _, pvc := range pvcs {
+			log.Printf("Found pvc %s in namespace %q\n", pvc.Name, pvc.Namespace)
+		}
+		By("Run crane export/transform/apply pipeline")
+		log.Printf("Running crane pipeline for namespace %s\n", srcApp.Namespace)
+		runner := scenario.Crane
```
msajidmansoori12: @nachandr this should be scenario.CraneNonAdmin to make sure we are running the tests as nonadmin/ns-admin user.

e2e-tests/tests/empty_pvc_migration_test.go
```diff
@@ -0,0 +1,188 @@
...
+				}
+			}
+
+			return nil
+		})
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Apply rendered manifests to target")
+		log.Printf("Applying rendered manifests on target namespace %s from %s\n", tgtApp.Namespace, paths.OutputDir)
+		Expect(kubectlTgtNonAdmin.ApplyDir(paths.OutputDir)).NotTo(HaveOccurred())
```
msajidmansoori12: @nachandr this applies to target cluster directly which bypasses the validation check , and also the addition you did `args := []string{"apply", "-R", "-f", dir, "--dry-run=server", "--warnings-as-errors=false"}`
this should be changed to use ApplyOutputToTargetNonAdmin() function which also does the validation dry run before actual apply.

msajidmansoori12: You can take a look at mtc_290_initcontainer_test.go file.

e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+			log.Printf("Transferring PVC %s to namespace %s on target cluster", pvcName, tgtApp.Namespace)
+			Expect(runner.TransferPVC(opts)).NotTo(HaveOccurred())
+			log.Printf("PVC transfer complete : %s -> namespace %s", pvcName, tgtApp.Namespace)
+		}
+
+		By("List pvcs on target cluster")
+		tgtpvcs, err := ListPVCs(tgtApp.Namespace, "", tgtApp.Context)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(tgtpvcs).NotTo(BeEmpty(), "expected at least one pvc in target namespace %q", tgtApp.Namespace)
```
msajidmansoori12: @nachandr Perhaps it would be good to also assert something like the PVC is actually empty.

e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"github.com/konveyor/crane/e2e-tests/config"
+	. "github.com/konveyor/crane/e2e-tests/framework"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Empty PVC migration", func() {
+	It("[MTC-153] Migrate an empty PVC associated with an application", Label("tier0"), func() {
+		appName := "app-with-empty-pvc"
```
msajidmansoori12, line 16: @nachandr  The CI is also failing because the application you are using is trying to look for pod.yml.j2 file which doesn't exist - https://github.com/stillalearner/k8s-apps-deployer/blob/main/k8sdeployer/ansible/roles/app-with-empty-pvc/tasks/deploy.yml#L15

e2e-tests/tests/empty_pvc_migration_test.go
msajidmansoori12: @nachandr I agree with marek this isn't needed , also since the CI runs on minikube clusters we don't specifically test for openshift , so should be removed from the test.

e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
```diff
@@ -0,0 +1,188 @@
...
+			opts := TransferPVCOptions{
+				SourceContext:   srcApp.Context,
+				TargetContext:   tgtApp.Context,
+				PVCName:         pvcName,
+				PVCNamespaceMap: fmt.Sprintf("%s:%s", srcApp.Namespace, tgtApp.Namespace),
+				Endpoint:        "route",
+				IngressClass:    "",
+				Subdomain:       "",
+			}
+			log.Printf("Transferring PVC %s to namespace %s on target cluster", pvcName, tgtApp.Namespace)
```
msajidmansoori12, line 111: @nachandr this also needs to change so that it works on minikube clusters on CI , as this may work on openshift but not on minikube setup that we use. You can take a look at `mtc_123_stateful_migration_test.go` to see what is needed.

msajidmansoori12 (changes requested):

e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
nachandr: That role has been updated through this PR : [Update app-with-empty-pvc](https://github.com/stillalearner/k8s-apps-deployer/pull/25)

msajidmansoori12: @nachandr left some comments on the PR please check.

e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
```diff
@@ -0,0 +1,139 @@
...
+				log.Printf("Removing temp dir: %s\n", paths.TempDir)
+				if err := os.RemoveAll(paths.TempDir); err != nil {
+					log.Printf("cleanup: failed to remove temp dir %q: %v", paths.TempDir, err)
+				}
+			}
+		})
+		DeferCleanup(func() {
+			By("Delete test namespace on source and target (best effort)")
+			for _, k := range []KubectlRunner{scenario.KubectlSrc, scenario.KubectlTgt} {
+				if _, err := k.Run("delete", "namespace", namespace, "--ignore-not-found=true", "--wait=false"); err != nil {
```
msajidmansoori12: @nachandr please change --wait=true this might cause accidental CI failures , we updated other tests yesterday as well.

msajidmansoori12 (changes requested):

e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
msajidmansoori12: Rest looks good to me

msajidmansoori12 (approved):
lgtm

Issue #188, opened 2026-03-24
Update transform and apply action with kustomize


aufi:
Details on https://github.com/migtools/crane/issues/207

PR #211, opened 2026-03-31
Update transform with kustomize and multistage

aufi:
There is still some work on polishing this feature to make it really helpful and stable (mostly in follow-ups), but its basic version should work and PR is ready for review now.

Sorry for bigger amount of changes in this PR, but I already put some additional sub-features away.

Related to https://github.com/migtools/crane/issues/188

## Summary by CodeRabbit

* **New Features**
  * Multi-stage Kustomize transform & apply workflows with stage selection, chaining, single/multi-stage execution, and --force overwrite
  * New CLI flags to select single, ranged, or list stage execution; explicit-flag detection drives behavior
  * Render manifests via kubectl kustomize and write per-stage or final outputs

* **Bug Fixes**
  * Validation enforces mutual exclusivity of stage-selection flags
  * Check for kubectl client availability before running transforms

* **Documentation**
  * Two new comprehensive guides on multi-stage Kustomize pipelines, usage, and migration

* **Tests**
  * Added fixtures and unit tests for stage discovery, filtering, navigation, path helpers, writers, and file helpers

* **Chores**
  * Output directory creation now uses stricter permissions (0700)


aufi:
@coderabbit review

internal/transform/orchestrator.go
```diff
@@ -0,0 +1,319 @@
...
+
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	o.Log.Debugf("Running: kubectl kustomize %s", stageDir)
+
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("failed to build stage %s: %w\nstderr: %s", stage.DirName, err, stderr.String())
+	}
```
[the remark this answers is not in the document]
aufi: kubectl kustomize doesn't call any external service here, if it hangs, it is OK to hang crane too instead of adding possibl short timeout (that would imply need for extension optionally).

cmd/apply/apply.go
```diff
@@ -83,11 +113,89 @@ func addFlagsForOptions(o *Flags, cmd *cobra.Command) {
...
+		}
+
+		log.Info("Applying selected stages...")
+		return applier.ApplyMultiStage(selector)
+	}
+
+	// Default or explicit --final-only: apply final stage
+	log.Info("Applying final stage...")
+	return applier.ApplyFinalStage()
+}
```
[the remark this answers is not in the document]
aufi: re-
check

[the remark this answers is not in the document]
aufi: tests added with https://github.com/migtools/crane/pull/211/commits/fde2ec2e474af25a28827f3ff0f6c42aac88f6d6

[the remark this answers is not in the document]
aufi: apply_test.go added, re-check

docs/TRANSFORM_README.md
```diff
@@ -0,0 +1,373 @@
+# Crane Transform Directory Structure
```
aufi, line 1: Docs files structure will be updated/unified with https://github.com/migtools/crane/issues/215

cmd/apply/apply.go
```diff
@@ -83,18 +100,16 @@ func addFlagsForOptions(o *Flags, cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&o.ExportDir, "export-dir", "e", "export", "The path where the kubernetes resources are saved")
```
msajidmansoori12, line 102: @aufi Hi , I see that --export-dir isn't being used but still getting registered with the CLI , I think it would be good to remove it , deferring it to your expertise if you think it should be taken care of in a follow up PR.

```
./marek-crane apply -h              
Apply the transformations to the exported resources and save results in an output directory

Usage:
  crane apply [flags]

Flags:
  -e, --export-dir string      The path where the kubernetes resources are saved (default "export")```

**Review by msajidmansoori12 — COMMENTED** [review]
@aufi Hi I noticed that in output dir manifests , the yaml indentation contains 4 spaces while in general the manifests we get from kubectl contains 2 spaces. Just a formatting difference not something that needs to be fixed right away just wanted to make a note of it.

```
apiVersion: apps/v1
kind: Deployment
metadata:
    annotations:
        deployment.kubernetes.io/revision: "1"
    labels:
        name: simple-nginx-nopv
    name: simple-nginx-nopv-deployment
    namespace: simple-nginxx-nopv
spec:
    progressDeadlineSeconds: 600
    replicas: 1
    revisionHistoryLimit: 10
    selector:
        matchLabels:
            app: simple-nginx-nopv
    strategy:
        rollingUpdate:
            maxSurge: 25%
            maxUnavailable: 25%
        type: RollingUpdate
    template:
```

File: internal/transform/orchestrator.go [hunk 5]
```diff
@@ -0,0 +1,319 @@
+package transform
```
stillalearner, line 1: @aufi It would be great if we have some unit tests for this file

cmd/transform/transform.go
```diff
@@ -96,6 +117,16 @@ func addFlagsForOptions(o *Flags, cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.IgnoredPatchesDir, "ignored-patches-dir", "", "The path where files that contain transformations that were discarded due to conflicts are saved. If left blank, these files will not be saved.")
 	cmd.Flags().StringSliceVar(&o.PluginPriorities, "plugin-priorities", nil, "A comma-separated list of plugin names. A plugin listed will take priority in the case of patch conflict over a plugin listed later in the list or over one not listed at all.")
 	cmd.Flags().StringVar(&o.OptionalFlags, "optional-flags", "", "JSON string holding flag value pairs to be passed to all plugins ran in transform operation. (ie. '{\"foo-flag\": \"foo-a=/data,foo-b=/data\", \"bar-flag\": \"bar-value\"}')")
+
+	// Multi-stage flags
+	cmd.Flags().StringVar(&o.Stage, "stage", "", "Run transform for a specific stage only (e.g., '10_kubernetes')")
```
stillalearner: @aufi The actual plugin name is KubernetesPlugin, So the default stage name should be 10_KubernetesPlugin

Tested locally with 10_kubernetes, in case of multistage, "kubernetes" plugin name does not match with "KubernetesPlugin" and hence doesnt work.

internal/apply/kustomize.go
```diff
@@ -0,0 +1,256 @@
...
+
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("command failed: %w\nstderr: %s", err, stderr.String())
+	}
+
+	return stdout.Bytes(), nil
+}
+
+// ValidateKubectlAvailable checks if kubectl command is available
+func ValidateKubectlAvailable() error {
```
msajidmansoori12, line 172: @aufi noticed that this function is defined but not used, perhaps it would be good to use it in run() in apply.go after the export , transformDir and outputDir paths are resolved , so that crane fails early if kubectl is not present rather than from kubectl kustomize later ?

Something like this maybe in cmd/apply/apply.go ?
```
func (o *Options) run() error {
	log := o.globalFlags.GetLogger()

	transformDir, err := filepath.Abs(o.TransformDir)
	if err != nil {
		return err
	}

	outputDir, err := filepath.Abs(o.OutputDir)
	if err != nil {
		return err
	}

	if err := apply.ValidateKubectlAvailable(); err != nil {
		return err
	}

	// Create output directory
	if err := os.MkdirAll(outputDir, 0700); err != nil {
		return fmt.Errorf("failed to create output directory: %w", err)
	}
```

cmd/apply/apply.go
aufi: <s>Thanks, export dir flag removed.</s>

internal/transform/orchestrator.go
aufi: Added, thanks!

internal/apply/kustomize.go
aufi: Thanks for finding it, added the Validate method call.

cmd/transform/transform.go
aufi: You're right, thanks, fixed.

cmd/apply/apply.go
aufi: Tests and original crane flow expected it (while current multistage&kustomize transform copy it to fit transformation to the data and to move reources across stages).

I put it back until examine other (compatible) options.

internal/transform/stages.go
```diff
@@ -0,0 +1,204 @@
...
+	PluginName string
+	DirName    string
+	Path       string
+}
+
+// DiscoverStages scans transform directory for stage subdirectories
+// and returns them sorted by priority (numeric prefix)
+func DiscoverStages(transformDir string) ([]Stage, error) {
+	// Pattern: <number>_<pluginName>
+	pattern := regexp.MustCompile(`^([0-9]+)_([a-zA-Z0-9_-]+)$`)
```
nachandr, line 24: The current stage name validation regex only allows [a-zA-Z0-9_-]+ for the plugin portion, which excludes dots (.).

This prevents valid Kubernetes-style identifiers like route.openshift.io from being used in stage names (e.g., 30_route.openshift.io would fail validation).

Considering that many Kubernetes API groups and resources use dotted naming conventions, it would be beneficial to support . in plugin names.

cmd/apply/apply.go
msajidmansoori12: Sounds good to me , thanks !

msajidmansoori12 (approved):

internal/transform/stages.go
aufi: Thanks for comment. Existing plugin names usually describe their use-case (typically with CamelCase describing technology or something - https://github.com/migtools/crane-plugins/tree/main/plugins, so in this case maybe  `RouteToGatewayPlugin` would work too).

I'd keep the regex validation without `.`, just might document it in some plugin development doc.

stillalearner (approved):
lgtm. Thanks @aufi , great PR

PR #256, opened 2026-04-14
Update kustomize transform multistage

### LINKED ISSUE #251: Update multistage transform logic
Current handling of between-stages results is not fully correct and in many cases buggy. This need some cleanup including mulstistage flow update, CLI flags simplification (any maybe make apply just an alias for something like kustomize --apply-only, since transform need execute apply after each stage to be able create transformation for following stage).

Multistage crane transform should always be sequentially consistent. Stage N+1 must consume the fully applied output of stage N, not the original export snapshot.
Otherwise stage ordering is only ordered patch generation over stale input, which breaks whiteouts, structural rewrites, and any plugin logic depending on prior changes. If Crane supports multistage, it should mean real state transitions between stages by definition.

It should work well with default plugins (KubernetesPlugin), custom plugins and custom kustomizations in multistage way keeping all steps resources in sync.

Expected flow to be supported:
1. crane transform (creates default plugin transformation)
2. crane transform --stage-name 20_OpenShiftPlugin --plugin-name OpenShiftPlugin (better without --plugin-name needed)
3. crane transform --stage-name 30_CustomModification
(anyway these commands might be revised to provide better/more logical UX)

Result:
1. all resources/ directories in stages are updated with up-to-date content
2. output.yaml and result resources/ exists and their content matches

aufi:
Providing update on crane transform and apply with kustomize. Primary fixing multi-stage flow and other fixes and removals of not needed flags (from/to stages selectors):

• correctness of stage-to-stage chaining, sequential approach
• whiteout propagation
• regressions in apply/output behavior
• behavior of custom pass-through stages, and their re-write

Including minor transform-related doc update.

Sorry for large PR, will do my best to create multiple smaller PRs next time.

Fixes https://github.com/migtools/crane/issues/251

## Summary by CodeRabbit

* **New Features**
  * CLI applies all discovered stages by default, supports running a single stage via --stage, auto-detects kubectl/oc, and writes final-stage output split into per-resource files with per-stage input/output snapshots for debugging.

* **Documentation**
  * Added a comprehensive multistage transform guide covering layout, commands, stage discovery/options, debugging/workdir guidance, and examples.

* **Bug Fixes**
  * Ensured sequential materialization of stages and correct exclusion of deleted/whiteout resources across stages.

* **Tests**
  * Added extensive end-to-end tests for multi-stage sequencing, whiteout behavior, working directories, and failure scenarios.


aufi:
@coderabbitai review

cmd/transform/transform.go
```diff
@@ -219,3 +238,58 @@ func optionalFlagsToLower(inFlags map[string]string) map[string]string {
...
+			log.Infof("Previous stage %s hasn't been run yet, running it first...", lastStage.DirName)
+
+			// Mark as newly created to allow population
+			orchestrator.NewlyCreatedStages[lastStage.DirName] = true
+
+			selector := internalTransform.StageSelector{Stage: lastStage.DirName}
+			if err := orchestrator.RunMultiStage(selector); err != nil {
+				return fmt.Errorf("failed to run previous stage %s: %w", lastStage.DirName, err)
+			}
+		}
```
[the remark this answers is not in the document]
aufi: re-check

docs/transform.md
```diff
@@ -0,0 +1,540 @@
...
+crane apply --transform-dir transform --output-dir output
+
+# Apply specific stage only
+crane apply --stage 10_KubernetesPlugin
+
+# Apply specific stage range
+crane apply --from-stage 10_KubernetesPlugin --to-stage 20_OpenshiftPlugin
+```
+
+**Note**: The default behavior applies **all stages** sequentially to ensure each stage output is properly materialized. This maintains sequential consistency across the entire pipeline.
```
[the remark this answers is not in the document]
aufi: Docs will be update in follow-up with more transform scenarios.

cmd/apply/apply.go
```diff
@@ -29,34 +29,16 @@ type Flags struct {
...
-	}
-	if len(o.Flags.Stages) > 0 {
-		flagCount++
-	}
-
-	if flagCount > 1 {
-		return fmt.Errorf("--stage, --from-stage/--to-stage, and --stages are mutually exclusive")
-	}
-
+	// No validation needed - only --stage flag exists
```
stillalearner: any validation needed to check if --stage value is correctly passed?

internal/apply/kustomize.go
stillalearner: Is it being used anymore?

internal/transform/orchestrator.go
```diff
@@ -121,48 +51,84 @@ func (o *Orchestrator) RunMultiStage(stageSelector StageSelector) error {
...
+
+		// Step 3: Save input to working directory for debugging
+		stageInputDir := opts.GetStageInputDir(stage.DirName)
+		if err := o.writeResourcesToDirectory(inputResources, stageInputDir); err != nil {
+			return fmt.Errorf("stage %s: failed to write input snapshot: %w", stage.DirName, err)
 		}
 
-		// Execute transform for this stage
+		// Step 4: Execute transform for this stage (generates transform artifacts)
 		if err := o.executeStage(stage, inputResources); err != nil {
```
stillalearner: it would be good to load all plugins once in RunMultiStage function at the top, and pass those plugins list here, so to avoid calling GetFilteredPlugins function everytime as each call to GetFilteredPlugins spawns a subprocess for every plugin binary in the directory 

`  func (o *Orchestrator) RunMultiStage(stageSelector StageSelector) error {                                                                                                                                                 
      stages, err := DiscoverStages(o.TransformDir)                                                                                                                                                                         
      ...                                                                                                                                                                                                                   
                                                                                                                                                                                                                            
      // Load plugins once for the entire pipeline                                                                                                                                                                          
      allPlugins, err := plugin.GetFilteredPlugins(o.PluginDir, o.SkipPlugins, o.Log)
      if err != nil {                                                                                                                                                                                                       
          return fmt.Errorf("failed to load plugins: %w", err)
      }                                                                                                                                                                                                                     
                            
      for i, stage := range selectedStages {
          ...
          if err := o.executeStage(stage, inputResources, allPlugins); err != nil {
              ...                                                                                                                                                                                                           
          }
      }                                                                                                                                                                                                                     
      ...                   
  }
                                                                                                                                                                                                                            
  func (o *Orchestrator) executeStage(stage Stage, inputResources []unstructured.Unstructured, allPlugins []cranelib.Plugin) error {
      stagePlugin := o.findPluginForStage(stage, allPlugins) // pure lookup, no I/O                                                                                                                                         
      ...                                                                                                                                                                                                                   
  }
                                                                                                                                                                                                                            
  func (o *Orchestrator) findPluginForStage(stage Stage, allPlugins []cranelib.Plugin) cranelib.Plugin {                                                                                                                    
      for _, p := range allPlugins {
          if p.Metadata().Name == stage.PluginName {                                                                                                                                                                        
              return p                                                                                                                                                                                                      
          }                                                                                                                                                                                                                 
      }                                                                                                                                                                                                                     
      return nil                                                                                                                                                                                                            
  }           `

cmd/apply/apply.go
aufi: Good point, there was originally just check for exclusive arguments vaiidation, added stage name validation for format.

internal/apply/kustomize.go
aufi: No, just in tests, removed, thanks.

internal/transform/orchestrator.go
aufi: Makes sense, updated.

internal/transform/writer.go
```diff
@@ -55,18 +59,54 @@ func (w *KustomizeWriter) WriteStage(artifacts []cranelib.TransformArtifact, for
...
+		resourceID := getResourceID(artifact.Resource)
+
+		// Check for duplicates
+		if _, exists := allResourcesMap[resourceID]; exists {
+			// Duplicate detected - determine which to keep
+			existingIsWhiteout := whiteoutStatusMap[resourceID]
+
+			if existingIsWhiteout && !artifact.HaveWhiteOut {
+				// Replace whiteout with non-whiteout
+				logrus.Warnf("Duplicate resource %s: replacing whiteout with active resource", resourceID)
```
stillalearner: can we have logger in KustomizeWriter instead of using global logger here

` type KustomizeWriter struct {                                                                                                                                                                                             
      opts      file.PathOpts
      stageName string                                                                                                                                                                                                      
      log       *logrus.Logger
  }                                                                                                                                                                                                                         
                            
  func NewKustomizeWriter(opts file.PathOpts, stageName string, log *logrus.Logger) *KustomizeWriter {                                                                                                                      
      return &KustomizeWriter{
          opts:      opts,                                                                                                                                                                                                  
          stageName: stageName,
          log:       log,
      }
  }`

internal/transform/writer.go
```diff
@@ -146,41 +211,17 @@ func (w *KustomizeWriter) WriteStage(artifacts []cranelib.TransformArtifact, for
...
-func capitalizeFirst(s string) string {
-	if s == "" {
-		return ""
-	}
-	// Simple capitalization - assumes ASCII
-	first := s[0]
-	if first >= 'a' && first <= 'z' {
-		return string(first-32) + s[1:]
+	if namespace != "" {
+		return fmt.Sprintf("%s/%s/%s", kind, namespace, name)
```
stillalearner, line 224: Can we use file.GetResourceFilename here, for filename consistency.

 A user debugging a failed migration might look at .work/10_KubernetesPlugin/output/default/ConfigMap_default_my-config.yaml and then try to find the same resource in 10_KubernetesPlugin/resources/ — but it's named     
  ConfigMap__v1_default_my-config.yaml there. The inconsistency is confusing.

internal/transform/writer.go
aufi: Agreed, good for consistency.

internal/transform/writer.go
aufi: Good point, updated to always use the shared method to determine resource filename.

docs/transform.md
```diff
@@ -0,0 +1,575 @@
...
+# Create a new plugin-based stage automatically
+# Stage name ending with "Plugin" will use that plugin
+crane transform --stage 35_OpenshiftPlugin
+
+# Create a new pass-through stage for manual editing
+# Stage name NOT ending with "Plugin" creates empty pass-through
+crane transform --stage 40_CustomManualEdits
+
+# Run from a specific stage onwards
+crane transform --from-stage 10_KubernetesPlugin
```
stillalearner: not valid any more?

docs/transform.md
```diff
@@ -0,0 +1,575 @@
...
+
+# Run specific stage range
+crane transform --from-stage 10_KubernetesPlugin --to-stage 20_OpenshiftPlugin
```
stillalearner: this also?

docs/transform.md
```diff
@@ -0,0 +1,575 @@
...
+
+# Run specific stages only
+crane transform --stages 10_KubernetesPlugin,30_ImagestreamPlugin
```
stillalearner: and this?

stillalearner (approved):
lgtm

Issue #286, opened 2026-04-20
[BUG] Variable shadowing bug in AsString() prevents errors from appearing in output


RanWurmbrand:
### Variable shadowing bug in AsString() prevents errors from appearing in output

**What version of crane are you running, and what are your clusters+platform**

Latest main branch. Issue exists in `cmd/transfer-pvc/progress.go` line 262.

**What did you expect to happen?**

When a PVC transfer completes with errors (`p.Errors` is non-empty), the `AsString()` method should include those errors in its output.

**What actually happened?**

Errors are never displayed. Line 262 uses `:=` instead of `=`, creating a new local variable `errors` that shadows the outer one declared on line 253. The inner variable gets populated then goes out of scope. The outer one stays empty and is returned.

**Please include any relevant logs or errors**

Fix: change line 262 from `errors :=` to `errors =`.

---

Suggested label: `good first issue`

PR #290, opened 2026-04-20
:test_tube: Add unit tests for Progress type in transfer-pvc

RanWurmbrand:
### Summary

- Adds comprehensive unit test coverage for the Progress type in `cmd/transfer-pvc/progress.go`
- Introduces `resetGlobals()` helper to isolate tests from shared global state (`pastAttempts`, `failedFiles`)

### What's covered

- `parseRsyncLogs`: parsing stdout/stderr from rsync, extracting transfer stats and failed files
- `NewProgress`: initialization with PVC name
- `Status()`: all status transitions (exit codes, percentages, completion states)
- `Merge()`: field merging, percentage aggregation across retries, capping at 100%, pastAttempts accumulation
- `AsString()`: output formatting with various field combinations
- `writeProgressToFile`: JSON output to file and error handling for invalid paths
  
### Also documents

- Variable shadowing bug on line 262 via skipped test `TestProgressAsString_WithErrors` — errors never appear in output because `:=` shadows the outer variable 
(see #286)

### Test plan

```
go test ./cmd/transfer-pvc/... -v -run "TestProgress|Test_parseRsyncLogs|TestWriteProgressToFile"
```

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for transfer progress initialization and aggregation.
  * Added validation for percentage updates, retry handling, data-size tracking, and error or failed-file collection.
  * Added coverage for saving progress as JSON, including invalid-path error handling.
  * Improved test isolation by resetting shared progress state between test cases.

### DIFF
--- a/cmd/transfer-pvc/progress_test.go
+++ b/cmd/transfer-pvc/progress_test.go
@@ -1,7 +1,9 @@
 package transfer_pvc
 
 import (
+	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 	"time"
@@ -687,5 +689,47 @@ func TestProgressAsString_WithFailedFiles(t *testing.T) {
 	}
 	if !strings.Contains(errStr, "Permission denied") {
 		t.Errorf("AsString() err output missing 'Permission denied'")
+	}
+}
+
+func TestWriteProgressToFile_WriteValidFile(t *testing.T) {
+	resetGlobals()
+	tmpPath := t.TempDir() + "/test-progress.json"
+
+	pct := int64(50)
+	p := &Progress{
+		TransferPercentage: &pct,
+		TransferredFiles:   100,
+		startedAt:          time.Now(),
+	}
+
+	err := writeProgressToFile(tmpPath, p)
+	if err != nil {
+		t.Errorf("writeProgressToFile() error = %v", err)
+	}
+
+	data, err := os.ReadFile(tmpPath)
+	if err != nil {
+		t.Errorf("Failed to read written file: %v", err)
+	}
+
+	var readProgress Progress
+	if err := json.Unmarshal(data, &readProgress); err != nil {
+		t.Errorf("Failed to unmarshal progress: %v", err)
+	}
+
+	if readProgress.TransferredFiles != 100 {
+		t.Errorf("Written TransferredFiles = %v, want 100", readProgress.TransferredFiles)
+	}
+}
+
+func TestWriteProgressToFile_InvalidPath(t *testing.T) {
+	resetGlobals()
+	p := &Progress{
+		TransferredFiles: 100,
+	}
+	err := writeProgressToFile("/nonexistent/directory/file.json", p)
+	if err == nil {
+		t.Errorf("writeProgressToFile() with invalid path should return error")
 	}
 }
--- a/cmd/transfer-pvc/progress_test.go
+++ b/cmd/transfer-pvc/progress_test.go
@@ -2,8 +2,18 @@ package transfer_pvc
 
 import (
 	"fmt"
+	"strings"
 	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/types"
 )
+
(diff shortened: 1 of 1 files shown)


cmd/transfer-pvc/progress_test.go
```diff
@@ -178,3 +190,546 @@ total size is 8.67G  speedup is 1.00`,
...
+	if p.TransferPercentage == nil {
+		t.Errorf("Merge() TransferPercentage = nil, want 50")
+		return
+	}
+	if *p.TransferPercentage != 50 {
+		t.Errorf("Merge() TransferPercentage = %d, want 50 (pastAttempts 30 + in 20)", *p.TransferPercentage)
+	}
+}
+
+func TestProgressMerge_PercentageCapAt100(t *testing.T) {
```
aufi, line 447: This function looks little weird to me, but seems to be in sync with the tested source code. We might keep in mind those tests are testing currently existing code as-is, not saying the code implementation is correct or no.

aufi (commented):
Thanks for the PR Ran! Overall looks good.

I'd suggest add `parseRsyncLogs`  https://github.com/migtools/crane/blob/main/cmd/transfer-pvc/progress.go#L399 method test. The methods seems quite important to be tested regarding to its logic. 

(I'm not sure about pod-related methods in the end of original file, but those might not be required)

cmd/transfer-pvc/progress_test.go
```diff
@@ -341,3 +369,296 @@ func TestStatus_NilTransferredData(t *testing.T) {
...
+func TestWriteProgressToFile_InvalidPath(t *testing.T) {
+	resetGlobals()
+	p := &Progress{
+		TransferredFiles: 100,
+	}
+	err := writeProgressToFile("/nonexistent/directory/file.json", p)
+	if err == nil {
+		t.Errorf("writeProgressToFile() with invalid path should return error")
+	}
+}
```
midays: Both the PR description and both commit messages say this PR adds tests for `AsString()`, including a specific claim: "Document variable shadowing bug on line 262 via skipped test" (referencing #286). I checked this diff and both individual commits' patches directly through the GitHub API — there's no `AsString` test anywhere, skipped or otherwise.

Separately: issue #286 (the shadowing bug this was meant to document) is already closed — it was fixed by #622, unrelated to this PR. I checked current `main` and `AsString()` now correctly uses `errors = "Errors: \n"` (no shadowing). So even if the `AsString()` test gets added now, framing it as documenting a known/skipped bug would be stale — the bug's already fixed on `main`. Could you either add real `AsString()` coverage (testing the now-correct behavior) or update the description/commit messages to drop this claim?

cmd/transfer-pvc/progress_test.go
```diff
@@ -341,3 +369,296 @@ func TestStatus_NilTransferredData(t *testing.T) {
...
+		Errors:             []string{},
+	}
+	in := &Progress{
+		TransferPercentage: &pct2,
+		FailedFiles:        []FailedFile{},
+		Errors:             []string{},
+	}
+	p.Merge(in)
+	if p.TransferPercentage == nil {
+		t.Errorf("Merge() TransferPercentage = nil, want non-nil")
```
midays: This only asserts `TransferPercentage != nil`, not the actual value. Given `Merge()`'s logic (pastAttempts + incoming, capped at 100, only updates if higher than current), this test would still pass even if `Merge()` computed a wrong percentage or did nothing at all to it, as long as the field stayed non-nil. The name suggests it's testing aggregation specifically, but `TestProgressMerge_PercentageBasicUpdate` and `_PercentageAccumulationWithPastAttempts` right below it already do this more rigorously (asserting exact values). Worth tightening this one to assert the expected value too, or dropping it if it's redundant with those.

midays (changes requested):

cmd/transfer-pvc/progress_test.go
RanWurmbrand: good catch! thanks

PR #304, opened 2026-04-22
Automate MTC-197: Add PVC data integrity test with MD5 verification

### LINKED ISSUE #303: Automate MTC-197 : Add PVC data integrity test with MD5 verification

nachandr:
Resolves [Issue#303](https://github.com/migtools/crane/issues/303)

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test that validates PVC data integrity during migration using non-admin contexts. It provisions source and target apps, generates a test file and checksum on the source, transfers PVCs to the target, applies rendered manifests, and verifies the migrated file and MD5 checksum match the source. The test skips if required contexts are missing and performs cleanup.
* **Tests**
  * Improved test helper behavior and standardized an error message to make failures clearer during test runs.

### DIFF
--- a/e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
+++ b/e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
@@ -119,6 +119,19 @@ var _ = Describe("Empty PVC migration", func() {
 		Expect(tgtpvcs).NotTo(BeEmpty(), "expected at least one PVC in target namespace %q", tgtApp.Namespace)
 		log.Printf("Found %d PVCs in target namespace %q", len(tgtpvcs), tgtApp.Namespace)
 
+		// Verify each source PVC was transferred to target
+		for _, srcPVC := range pvcs {
+			found := false
+			for _, tgtPVC := range tgtpvcs {
+				if tgtPVC.Name == srcPVC.Name {
+					found = true
+					log.Printf("Verified PVC %s exists on target\n", srcPVC.Name)
+					break
+				}
+			}
+			Expect(found).To(BeTrue(), "expected source PVC %q to exist on target", srcPVC.Name)
+		}
+
 		By("Apply rendered manifests to target")
 		log.Printf("Applying rendered manifests on target namespace %s from %s\n", tgtApp.Namespace, paths.OutputDir)
 		Expect(ApplyOutputToTargetNonAdmin(kubectlTgtNonAdmin, paths.OutputDir)).NotTo(HaveOccurred())
--- a/e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
+++ b/e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
@@ -119,19 +119,6 @@ var _ = Describe("Empty PVC migration", func() {
 		Expect(tgtpvcs).NotTo(BeEmpty(), "expected at least one PVC in target namespace %q", tgtApp.Namespace)
 		log.Printf("Found %d PVCs in target namespace %q", len(tgtpvcs), tgtApp.Namespace)
 
-		// Verify each source PVC was transferred to target
-		for _, srcPVC := range pvcs {
-			found := false
-			for _, tgtPVC := range tgtpvcs {
-				if tgtPVC.Name == srcPVC.Name {
-					found = true
-					log.Printf("Verified PVC %s exists on target\n", srcPVC.Name)
-					break
-				}
-			}
-			Expect(found).To(BeTrue(), "expected source PVC %q to exist on target", srcPVC.Name)
-		}
-
 		By("Apply rendered manifests to target")
 		log.Printf("Applying rendered manifests on target namespace %s from %s\n", tgtApp.Namespace, paths.OutputDir)
 		Expect(ApplyOutputToTargetNonAdmin(kubectlTgtNonAdmin, paths.OutputDir)).NotTo(HaveOccurred())
--- a/e2e-tests/tests/mtc_197_pvc_data_integrity_test.go
+++ b/e2e-tests/tests/mtc_197_pvc_data_integrity_test.go
@@ -1,0 +1,161 @@
+package e2e
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/konveyor/crane/e2e-tests/config"
+	. "github.com/konveyor/crane/e2e-tests/framework"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("PVC data integrity migration", func() {
+	It("[MTC-197] Migrate a PVC with data and verify checksum integrity", Label("tier0"), func() {
+		appName := "app-with-empty-pvc"
+		namespace := appName
+		scenario := NewMigrationScenario(
+			"app-with-empty-pvc",
+			namespace,
+			config.K8sDeployBin,
+			config.CraneBin,
+			config.SourceContext,
+			config.TargetContext,
+		)
+		if scenario.KubectlSrcNonAdmin.Context == "" {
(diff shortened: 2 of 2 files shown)


nachandr:
> **Actionable comments posted: 1**
> 
> 🧹 Nitpick comments (2)
> > e2e-tests/tests/mtc_197_pvc_data_integrity_test.go (2)> `132-136`: **`tgtpvcs` is only length-checked — consider verifying PVC names match.**
> > > Since `TransferPVC` is invoked per-source-PVC, asserting that each transferred PVC name exists on the target (rather than only `len > 0`) would catch partial-transfer regressions. Optional.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@e2e-tests/tests/mtc_197_pvc_data_integrity_test.go` around lines 132 - 136,
> > > The test currently only checks tgtpvcs length after calling ListPVCs; update the
> > > assertion to verify that each source PVC transferred by TransferPVC exists by
> > > name in the returned tgtpvcs slice. Locate where TransferPVC is invoked and
> > > capture the source PVC names (or the source PVC object list), then after calling
> > > ListPVCs(tgtApp.Namespace, "", tgtApp.Context) assert that for every expected
> > > source PVC name there is a matching PVC in tgtpvcs (compare Name fields),
> > > failing the test if any expected name is missing; use the existing variables
> > > tgtpvcs, tgtApp.Namespace and the same ListPVCs call to perform the name
> > > presence checks.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > `148-156`: **Target-side exec relies on `Validate()` fully readying the pod.**
> > > The `Eventually(tgtApp.Validate, "2m", "10s")` at Line 144 should guarantee the pod is ready before the subsequent `kubectl exec` calls. If `Validate()` ever becomes less strict (e.g., returns before the PVC is mounted or the pod is Running), Lines 148/153/159 would flake. Consider wrapping the exec/MD5 verification in an `Eventually` block as a defensive measure.

> > > 
> > > In `@e2e-tests/tests/mtc_197_pvc_data_integrity_test.go` around lines 148 - 156,
> > > The pod readiness/volume mount calls to kubectlTgtNonAdmin.Run (used for ls and
> > > md5 verification) can flake if tgtApp.Validate becomes less strict; wrap the
> > > execs that run ls /data/<fileName> and the md5 verification command (the
> > > kubectlTgtNonAdmin.Run invocations that reference appName and fileName and the
> > > md5sum -c call) inside an Eventually that retries until the pod truly has the
> > > file and PVC mounted (e.g., Eventually(func() { <run ls and md5 via
> > > kubectlTgtNonAdmin.Run and assert no error and contains "OK"> }, "2m", "10s")),
> > > so the checks only proceed once the pod/pvc are fully ready.

validate.yml already has the retry logic with pod readiness checks.

e2e-tests/tests/mtc_153_empty_pvc_migration_test.go
```diff
@@ -119,6 +119,19 @@ var _ = Describe("Empty PVC migration", func() {
...
+			found := false
+			for _, tgtPVC := range tgtpvcs {
+				if tgtPVC.Name == srcPVC.Name {
+					found = true
+					log.Printf("Verified PVC %s exists on target\n", srcPVC.Name)
+					break
+				}
+			}
+			Expect(found).To(BeTrue(), "expected source PVC %q to exist on target", srcPVC.Name)
+		}
```
msajidmansoori12: @nachandr perhaps the function you added can be used here `func VerifyPVCsExistByName()`

e2e-tests/framework/client.go
```diff
@@ -36,6 +36,24 @@ func ListPVCs(namespace string, labelSelector string, contextName string) ([]cor
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("source PVC %q not found in target PVCs", srcPVC.Name)
+		}
+	}
+	return nil
+}
```
msajidmansoori12, line 60: @nachandr just a suggestion , currently i can see the function only returns the first missing PVC which makes triaging a bit noisier and we may have to run this function multiple times to check for the next missing PVC, i think a better option to explore would be to return an error with all the missing PVC names.

Deferring it to you.

e2e-tests/tests/mtc_197_pvc_data_integrity_test.go
```diff
@@ -0,0 +1,172 @@
...
+			By("Cleanup temp directory")
+			if paths.TempDir != "" {
+				log.Printf("Removing temp dir: %s\n", paths.TempDir)
+				if err := os.RemoveAll(paths.TempDir); err != nil {
+					log.Printf("cleanup: failed to remove temp dir %q: %v", paths.TempDir, err)
+				}
+			}
+		})
+		DeferCleanup(func() {
+			By("Delete test namespace on source and target (best effort)")
```
msajidmansoori12: `(best effort)` looks misleading since we will wait until the ns is deleted. Better option is keep  `(wait for completion)` instead as other tests used that too.

e2e-tests/tests/mtc_197_pvc_data_integrity_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+		By("Grant ns admin permissions to nonadmin user on source and target")
+		kubectlSrcNonAdmin, kubectlTgtNonAdmin, cleanup, err := SetupNamespaceAdminUsersForScenario(scenario, namespace)
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(cleanup)
+
+		By("SOURCE: Deploy and validate app")
+		log.Printf("Deploying source app %s in namespace %s\n", srcApp.Name, srcApp.Namespace)
+		Expect(srcApp.Deploy()).NotTo(HaveOccurred())
+		Expect(srcApp.Validate()).NotTo(HaveOccurred())
```
msajidmansoori12: @nachandr instead of using these manually a better option is to use the function `PrepareSourceAppNoQuiesce(srcApp)` also used by other tests.

e2e-tests/tests/mtc_197_pvc_data_integrity_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+		By("SOURCE: Get file MD5 checksum")
+		srcMD5Output, err := kubectlSrcNonAdmin.Run("exec", appName, "-n", srcApp.Namespace, "--", "/bin/sh", "-c", fmt.Sprintf("cat /data/%s.md5", fileName))
+		Expect(err).NotTo(HaveOccurred())
+		srcMD5 := strings.TrimSpace(srcMD5Output)
+		Expect(srcMD5).NotTo(BeEmpty(), "expected MD5 checksum file to exist on source")
+		log.Printf("Source: MD5 checksum: %s\n", srcMD5)
+
+		By("SOURCE: Quiesce app")
+		Expect(kubectlSrcNonAdmin.ScaleDeploymentIfPresent(srcApp.Namespace, srcApp.Name, 0)).NotTo(HaveOccurred())
```
msajidmansoori12: @nachandr since the application doesn't have a deployment  , this won't do anything actually so either remove this section , or please update the By ( ) as it can be misleading, atleast to something like `By("SOURCE: Quiesce app workload if deployment exists")` so that we know the app may or may not quiesce anything.

msajidmansoori12 (changes requested):
Added some suggestions, please review and make the changes.

e2e-tests/tests/mtc_197_pvc_data_integrity_test.go
nachandr: Thanks @msajidmansoori12

msajidmansoori12 (approved):

PR #321, opened 2026-04-30
Added support for offline validation without a live cluster availability

stillalearner:
## Summary                                                                                                                                                                                                                
   
  Adds offline GVK validation to `crane validate` via a new `--api-resources` flag. This enables validation against a target cluster's API surface without a live kubeconfig connection — supporting air-gapped             
  environments, CI/CD pipelines, and team handoff scenarios.          
                                                                                                                                                                                                                            
  ## Problem                                                          

  `crane validate` required a live connection to the target cluster. This blocked validation in disconnected environments and pipelines without cluster access.                                                             
   
  ## Solution                                                                                                                                                                                                               
                                                                      
  A capture script (`scripts/capture-api-surface.sh`) that calls `kubectl get --raw /apis/<group>/<version>` for every served group-version — producing the exact same data as `ServerGroupsAndResources()`. This eliminates
   the preferred-vs-all-served gap entirely.
                                                                                                                                                                                                                            
  **On target cluster** (only kubectl needed, no crane):              

      bash scripts/capture-api-surface.sh --context target-cluster -o api-surface.json

  **On migration workstation** (no cluster access needed):
                                                                                                                                                                                                                            
      crane validate -i output/ --api-resources api-surface.json
                                                                                                                                                                                                                            
  The script output is an array of native Kubernetes `APIResourceList` objects — the same type `ServerGroupsAndResources()` returns in Go. The parser uses `metav1.APIResourceList` directly with zero custom types or
  parsing heuristics.                                                                                                                                                                                                       
   
  `--api-resources` is mutually exclusive with `--context`/`--kubeconfig`. When neither is set, existing default kubeconfig behavior is unchanged.                                                                          
                                                                      
  ## Design decisions                                                                                                                                                                                                       
   
  - **Full API surface, not just preferred:** The capture script uses `kubectl api-versions` + `kubectl get --raw` to dump all served group-versions, matching the live path's `ServerGroupsAndResources()`. This was       
  changed after aufi's review identified that `kubectl api-resources -o json` only returns preferred versions.
  - **Subresource filtering parity:** Both live (`buildDiscoveryIndex`) and offline (`ParseAPIResourcesJSON`) skip resources with `/` in the name (e.g., `deployments/status`), ensuring consistent behavior.               
  - **No custom format:** The capture script output uses native Kubernetes `APIResourceList` JSON — no crane-specific types.                                                                                                
  - **Refactor, not rewrite:** `MatchResults` delegates to `MatchResultsFromIndex`. Both paths produce the same `DiscoveryIndex` type, so all matching logic is untouched.                                                  
                                                                                                                                                                                                                            
  ## Test plan                                                                                                                                                                                                              
                                                                                                                                                                                                                            
  - [x] `go build ./...` compiles                                     
  - [x] All unit tests pass (`go test ./internal/validate/... ./cmd/validate/...`)
  - [x] Existing live-mode tests unchanged and passing                                                                                                                                                                      
  - [x] `bash scripts/capture-api-surface.sh --context tgt -o api-surface.json` captures 27 API versions                                                                                                                    
  - [x] `crane validate -i output/ --api-resources api-surface.json` works in offline mode                                                                                                                                  
  - [x] `crane validate --api-resources api-surface.json --context foo` returns mutual exclusion error                                                                                                                      
  - [x] Cross-mode equivalence: 28 GVKs across 13 API groups tested — live and offline produce **identical** results (status, resourcePlural, suggestions all match)
  - [x] Subresource filtering consistent: `apps/v1 Scale` reported as "kind not found" in both modes

### DIFF
--- a/cmd/validate/validate.go
+++ b/cmd/validate/validate.go
@@ -19,15 +19,20 @@ type ValidateOptions struct {
 	cobraGlobalFlags *flags.GlobalFlags
 	globalFlags      *flags.GlobalFlags
 
-	inputDir     string
-	validateDir  string
-	outputFormat string
+	inputDir         string
+	validateDir      string
+	outputFormat     string
+	apiResourcesFile string
 
 	genericclioptions.IOStreams
 }
 
 // Complete loads the kubeconfig to ensure the target cluster is reachable.
+// Skipped in offline mode (--api-resources).
 func (o *ValidateOptions) Complete(c *cobra.Command, args []string) error {
+	if o.apiResourcesFile != "" {
+		return nil
+	}
 	_, err := o.configFlags.ToRawKubeConfigLoader().RawConfig()
 	if err != nil {
 		return fmt.Errorf("loading kubeconfig for target cluster: %w", err)
@@ -49,6 +54,18 @@ func (o *ValidateOptions) Validate() error {
 		return fmt.Errorf("--output must be \"yaml\" or \"json\", got %q", o.outputFormat)
 	}
 
+	if o.apiResourcesFile != "" {
+		if o.configFlags.Context != nil && *o.configFlags.Context != "" {
+			return fmt.Errorf("--api-resources and --context are mutually exclusive; use --api-resources for offline validation or --context for live validation")
+		}
+		if o.configFlags.KubeConfig != nil && *o.configFlags.KubeConfig != "" {
+			return fmt.Errorf("--api-resources and --kubeconfig are mutually exclusive; use --api-resources for offline validation or --kubeconfig for live validation")
+		}
+		if _, err := os.Stat(o.apiResourcesFile); err != nil {
+			return fmt.Errorf("api-resources file %q: %w", o.apiResourcesFile, err)
+		}
+	}
+
 	return nil
 }
 
@@ -63,15 +80,31 @@ func (o *ValidateOptions) Run() error {
 
 	log.Infof("Scanned %d distinct GVK+namespace tuples", len(entries))
 
-	discoveryClient, err := o.configFlags.ToDiscoveryClient()
-	if err != nil {
-		return fmt.Errorf("creating discovery client: %w", err)
-	}
-	discoveryClient.Invalidate()
+	var report *internalValidate.ValidationReport
 
-	report, err := internalValidate.MatchResults(entries, internalValidate.MatchOptions{DiscoveryClient: discoveryClient}, log)
-	if err != nil {
-		return fmt.Errorf("matching against target cluster: %w", err)
+	if o.apiResourcesFile != "" {
+		index, err := internalValidate.ParseAPIResourcesJSON(o.apiResourcesFile)
+		if err != nil {
+			return fmt.Errorf("loading api-resources: %w", err)
+		}
+		report = internalValidate.MatchResultsFromIndex(entries, index)
+		report.Mode = "offline"
+		report.APIResourcesSource = o.apiResourcesFile
+	} else {
+		discoveryClient, err := o.configFlags.ToDiscoveryClient()
+		if err != nil {
+			return fmt.Errorf("creating discovery client: %w", err)
+		}
+		discoveryClient.Invalidate()
+
+		report, err = internalValidate.MatchResults(entries, internalValidate.MatchOptions{DiscoveryClient: discoveryClient}, log)
+		if err != nil {
+			return fmt.Errorf("matching against target cluster: %w", err)
+		}
+		report.Mode = "live"
+		if o.configFlags.Context != nil && *o.configFlags.Context != "" {
+			report.ClusterContext = *o.configFlags.Context
+		}
 	}
 
 	internalValidate.FormatTable(o.Out, report)
@@ -129,6 +162,10 @@ GVK matching).
 
 Pipeline: export → transform → apply → validate
 
+Use --api-resources to validate offline against the JSON output of
+'kubectl api-resources -o json' when the target cluster is not directly
+reachable. Otherwise, supply kubeconfig/context flags for live validation.
+
 Incompatible resources are written to a failures/ directory under
 the validate-dir for auditability.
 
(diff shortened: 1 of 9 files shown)


cmd/validate/validate.go
```diff
@@ -19,15 +19,20 @@ type ValidateOptions struct {
...
+	outputFormat     string
+	apiResourcesFile string
 
 	genericclioptions.IOStreams
 }
 
 // Complete loads the kubeconfig to ensure the target cluster is reachable.
+// Skipped in offline mode (--api-resources).
 func (o *ValidateOptions) Complete(c *cobra.Command, args []string) error {
+	if o.apiResourcesFile != "" {
```
nachandr, line 33: Suggestion: Add api-resource file check in Complete() method

 Currently, when APIResourcesPath is set, Complete() returns early without verifying the file is accessible:
 ```                                                                                                                                                                    
  func (o *ValidateOptions) Complete(c *cobra.Command, args []string) error {
      if o.APIResourcesPath != "" {                                                                                                                                     
          return nil  // ← No validation of file accessibility
      }                                                                                                                                                                 
      // ... kubeconfig loading for live mode
  }                                                                                                                                                                     
```
While the file is checked later in Validate(), it would be better to verify file accessibility in Complete() for early failure detection. 

Suggested change:
```
  func (o *ValidateOptions) Complete(c *cobra.Command, args []string) error {
      // Offline mode: verify API resources file is accessible
      if o.APIResourcesPath != "" {
          if _, err := os.Stat(o.APIResourcesPath); err != nil {
              return fmt.Errorf("api-resources file not accessible: %w", err)
          }
          return nil
      }

      // Live mode: verify kubeconfig is accessible
      _, err := o.configFlags.ToRawKubeConfigLoader().RawConfig()
      if err != nil {
          return fmt.Errorf("loading kubeconfig for target cluster: %w", err)
      }
      return nil
  }
```

**Review by aufi — COMMENTED** [review]
Thanks for the PR Sachin! The offline validation use case makes sense and the refactor around a shared DiscoveryIndex is a nice direction.

I think there is one important correctness issue to address before this can be merged:

The new offline path is documented as consuming `kubectl api-resources -o json`, but that output does not represent the same API surface as the current live validation path.

The live path builds the discovery index from `ServerGroupsAndResources()`, which includes all served group/versions. In contrast, `kubectl api-resources -o json` is based on preferred resources (`ServerPreferredResources()` in kubectl). That means offline validation may incorrectly mark manifests using non-preferred but still served API versions as incompatible, even though live validation would accept them.

So the offline and live modes can produce different results for the same cluster.

I think we should either:
- change the offline input format/source so it contains all served APIResourceLists, matching `ServerGroupsAndResources()`, or
- clearly scope offline validation to preferred resources only and document that limitation.

There is also a smaller consistency issue: the live discovery index skips subresources (`strings.Contains(res.Name, "/")`), but `ParseAPIResourcesJSON` currently does not. The offline parser should probably apply the same filter so subresources cannot be treated as valid top-level manifest kinds.

(partially created by AI based on my&bots observations)

File: cmd/validate/validate.go (hunk 1 above)
Comment (stillalearner) [reply cmd/validate/validate.go:33]: Good suggestion, but the check is already in Validate() which runs right after Complete(). Moving it would duplicate the validation.

**Comment by stillalearner** [comment]
> Thanks for the PR Sachin! The offline validation use case makes sense and the refactor around a shared DiscoveryIndex is a nice direction.
> 
> I think there is one important correctness issue to address before this can be merged:
> 
> The new offline path is documented as consuming `kubectl api-resources -o json`, but that output does not represent the same API surface as the current live validation path.
> 
> The live path builds the discovery index from `ServerGroupsAndResources()`, which includes all served group/versions. In contrast, `kubectl api-resources -o json` is based on preferred resources (`ServerPreferredResources()` in kubectl). That means offline validation may incorrectly mark manifests using non-preferred but still served API versions as incompatible, even though live validation would accept them.
> 
> So the offline and live modes can produce different results for the same cluster.
> 
> I think we should either:
> 
> * change the offline input format/source so it contains all served APIResourceLists, matching `ServerGroupsAndResources()`, or
> * clearly scope offline validation to preferred resources only and document that limitation.
> 
> There is also a smaller consistency issue: the live discovery index skips subresources (`strings.Contains(res.Name, "/")`), but `ParseAPIResourcesJSON` currently does not. The offline parser should probably apply the same filter so subresources cannot be treated as valid top-level manifest kinds.
> 
> (partially created by AI based on my&bots observations)

Changes pushed as discussed. Please review.

**Review by aufi — APPROVED** [review]
LGTM

**Comment by aufi** [comment]
Adding a suggestion for possible docs update with this script functionality

To put the script logic into documentation, its content could be simplified. The key part is querying via `kubectl get --raw` for each API version. Here are the options:

  ## Oneliner (without parameters):
  ```bash
  kubectl api-versions | while read gv; do kubectl get --raw $([ "$gv" = "v1" ] && echo "/api/v1" || echo "/apis/$gv") 2>/dev/null || true; done | jq -s '{"apiResourceLists":.}' > api-surface.json
```
###  With --context and --kubeconfig support:
(this likely not for docs)
```
  KUBECTL_FLAGS="${1:+--context=$1} ${2:+--kubeconfig=$2}" OUTPUT="${3:-api-surface.json}"; kubectl api-versions $KUBECTL_FLAGS | while read gv; do kubectl get --raw $([ "$gv" = "v1" ] && echo "/api/v1" || echo "/apis/$gv") $KUBECTL_FLAGS 2>/dev/null || true; done | jq -s
  '{"apiResourceLists":.}' > "$OUTPUT"
```
 ## Recommendation for documentation:

  For documentation, I'd use the most readable version - a shortened script:

```
  #!/bin/bash
  # Usage: capture-api-surface.sh [-o output.json] [--context name] [--kubeconfig path]
  OUTPUT="api-surface.json"
  KUBECTL_FLAGS=""

  while [[ $# -gt 0 ]]; do
    case "$1" in
      -o) OUTPUT="$2"; shift 2 ;;
      --context|--kubeconfig) KUBECTL_FLAGS="$KUBECTL_FLAGS $1=$2"; shift 2 ;;
      *) echo "Unknown flag: $1"; exit 1 ;;
    esac
  done

  kubectl api-versions $KUBECTL_FLAGS | while read gv; do
    endpoint=$([ "$gv" = "v1" ] && echo "/api/v1" || echo "/apis/$gv")
    kubectl get --raw "$endpoint" $KUBECTL_FLAGS 2>/dev/null || true
  done | jq -s '{"apiResourceLists":.}' > "$OUTPUT"

  ```
The oneliner is shortest, but the shortened script is more readable and easier to copy/modify.

PR #358, opened 2026-05-07
Add --kustomize-args for transform and apply commands

### LINKED ISSUE #351: Add kustomize args option to transform and apply
Crane transform and apply uses `kustomize` tool internally, when user create their own custom stage, they might need adjust the kustomize evaluation with an additional arguments e.g. to allow multiple directory access or enable non-default features like `helm`.

This should be possible done by passing a new argument to the CLI (primary for transform command) and putting those argument to a file within the stage folder is to be considered (to not require add such arg to the apply command).

aufi:
This PR adds support for passing custom arguments to kubectl/oc kustomize via the new --kustomize-args flag in both transform and apply commands. Users can now enable advanced kustomize features like Helm chart support (--enable-helm), alpha plugins (--enable-alpha-plugins), custom load restrictions (--load-restrictor), and environment variables (--env). For now, kustomize extra args are not persisted within the transform stages files, but need to be passed in command args.

The implementation includes a security layer with argument whitelisting and injection prevention to safely validate and pass arguments to the underlying kustomize command.

Additionally, the deprecated runfn subcommand has been removed as its functionality can be achieved directly through kustomize itself.

Fixes https://github.com/migtools/crane/issues/351

## Summary by CodeRabbit

* **New Features**
  * Added a --kustomize-args option to apply and transform commands to customize kustomize invocation
  * Added a plugin manager top-level command

* **Removed Features**
  * Removed the runfn command

* **Tests**
  * Added unit tests for kustomize-args parsing and validation to ensure secure, well-formed arguments

[Review Change Stack](https://app.coderabbit.ai/change-stack/migtools/crane/pull/358)


internal/kustomize/args.go
```diff
@@ -0,0 +1,105 @@
...
+	"--enable-helm":          true,
+	"--env":                  true,
+	"-e":                     true,
+	"--helm-command":         true,
+	"--load-restrictor":      true,
+}
+
+// ParseAndValidateArgs parses and validates kustomize arguments
+// Returns array of arguments ready for exec.Command
+func ParseAndValidateArgs(argsString string) ([]string, error) {
```
aufi, line 20: There is a quite strict parsing of allowed content (keys are above) of the args (with tests below), I was not sure if do this strict&limited and hopefuly safe way or just check, that args don't escape the original command, so went this more safe way, but I'm open discuss it if needed.

internal/kustomize/args_test.go
```diff
@@ -0,0 +1,218 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := ParseAndValidateArgs(tt.input)
+
+			if tt.expectError {
+				if err == nil {
+					t.Errorf("expected error but got none")
+					return
+				}
+				if tt.errorMsg != "" && !contains(err.Error(), tt.errorMsg) {
```
midays: Is there a reason why a custome `contains()` function is implemented and used instead of the `string.contains` like it's used in [args.go](https://github.com/migtools/crane/pull/358/changes#diff-be83b728714017c84163d29596151e073073f9d42a6cd4dea08a259823d8b2b9R29) ?

```suggestion
				if tt.errorMsg != "" && !strings.Contains(err.Error(), tt.errorMsg) {
```

midays (commented):
LGTM, minor comment

internal/kustomize/args_test.go
aufi: You're right, thanks! Added another commit with suggestion above and few related changes.

internal/kustomize/args.go
```diff
@@ -0,0 +1,105 @@
...
+			argName = parts[0]
+		}
+
+		// Check whitelist
+		if !AllowedKustomizeArgs[argName] {
+			return nil, fmt.Errorf("kustomize argument %q is not allowed (security restriction)", argName)
+		}
+
+		// For value-taking flags, skip the next argument if it's space-separated
+		if valueTakingFlags[argName] {
```
msajidmansoori12, line 71: Hi @aufi , ParseAndValidateArgs() currently accepts malformed value-taking flags because it skips the next token without validating it.
This allows cases like --env --enable-helm to pass parsing (treating --enable-helm as env value) instead of failing early.
We should require a value for value-taking flags and reject when the next token is another flag.
Combinations i tried - 
```
Malformed / should fail early
--env
--env --enable-helm
--helm-command --enable-helm
--load-restrictor --enable-helm```

Hope this makes sense.

**Review by msajidmansoori12 — CHANGES_REQUESTED** [review]

File: internal/kustomize/args.go (hunk 3 above)
Comment (aufi) [reply internal/kustomize/args.go:71]: Thanks Sajid, updated!

File: internal/kustomize/args.go [hunk 4]
```diff
@@ -0,0 +1,118 @@
...
+		case r == ' ' && !inQuote:
+			if current.Len() > 0 {
+				args = append(args, current.String())
+				current.Reset()
+			}
+		default:
+			current.WriteRune(r)
+		}
+	}
+
```
msajidmansoori12, line 116: @aufi 
I think we should tighten quote validation in --kustomize-args parsing. Right now if someone passes an unclosed quote (for example --env 'FOO=bar), parsing still succeeds instead of returning an error. I reproduced this by adding a test case in TestParseAndValidateArgs expecting an error, and it fails with expected error but got none.

Can we add an explicit check for unclosed quotes at end-of-input and return a clear error (something like “unclosed quote in kustomize arguments”)?
Also please add regression tests for:

--env 'FOO=bar (unclosed single quote)
--env "FOO=bar (unclosed double quote)

msajidmansoori12 (changes requested):

internal/kustomize/args.go
aufi: Thanks, updated!

msajidmansoori12 (approved):
/lgtm , thanks for accomodating the comments

PR #361, opened 2026-05-09
Skip export of operator managed CRDs

stillalearner:
## Summary

  Skip export of CRDs that are managed by an operator. Operator-managed CRDs should be installed on the target cluster via their operator, not exported and applied directly — doing so creates orphaned CRDs that conflict
  when the operator is later installed.

  ## What changed
  
  Added `getOperatorManager()` in `cmd/export/crd.go` that checks four well-known operator markers on each CRD before exporting:
  
  - `olm.managed=true` label (OLM on OpenShift and K8s)
  - `app.kubernetes.io/managed-by` label (Helm, operator-sdk, ArgoCD, etc.)
  - `operators.operatorframework.io/*` annotations (operator-framework)
  - `ownerReferences` (CRD owned by a ClusterServiceVersion or operator Deployment)
  
  If any marker is detected, the CRD is skipped with a warning:
  
      Skipping CRD widgets.example.com — managed by operator cert-manager (OLM); install the operator on the target cluster instead

  CR instances are still exported normally — only the CRD definition is skipped. Unmanaged (standalone) CRDs continue to be exported as before.

  ## Test plan
  
  - [x] `go build ./...`
  - [x] All existing CRD export tests pass (7 tests)
  - [x] 5 new tests: skip OLM-managed, skip managed-by label, skip operator-framework annotation, skip owner-referenced, export unmanaged CRD
  - [x] 7 new `TestGetOperatorManager` sub-tests covering all marker combinations including negative cases (`olm.managed=false`, unrelated labels)
  - [x] Full unit test suite passes (`go test ./...` excluding e2e)

### DIFF
--- a/cmd/export/crd.go
+++ b/cmd/export/crd.go
@@ -51,6 +51,36 @@ func crdFailureAPIResourceName(crdName string) string {
 	return "customresourcedefinition-" + strings.ReplaceAll(crdName, "/", "-")
 }
 
+// getOperatorManager checks if a CRD is managed by an operator.
+// Returns the manager name if detected, empty string otherwise.
+func getOperatorManager(obj *unstructured.Unstructured) string {
+	labels := obj.GetLabels()
+	annotations := obj.GetAnnotations()
+
+	if v, ok := labels["olm.managed"]; ok && v == "true" {
+		if name := labels["operators.coreos.com/managed-by"]; name != "" {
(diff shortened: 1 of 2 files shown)


aufi (approved):
LGTM

midays (approved):
LGTM

PR #362, opened 2026-05-11
Added multiple olm operators test

stillalearner:
## Summary
  - Added E2E test for multiple OLM operators in the same namespace (tier1)
  - Deploys cert-manager and cockroachdb via shared CatalogSource, waits for both CSVs and InstallPlan, runs full crane pipeline
  - Verifies all OLM kinds from both operators are whiteout-ed with no partial leakage
  - Added `e2e-tests/testdata/olm-multi-operator-resources.yaml` with 2 Subscriptions, shared OperatorGroup and CatalogSource

  ## Test plan
  - [x] `go build ./...` passes
  - [x] `go vet ./e2e-tests/...` passes
  - [x] Manually verified on fresh minikube (`src` cluster with OLM v0.28.0): both CSVs created, all 3 assertions pass against real pipeline output

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end test data and a new "Multiple OLM operators" test suite to validate multi-operator deployment scenarios.
  * Test verifies export/transform/apply pipeline handles OLM resources: produces whiteout files for OLM kinds, generates exactly two Subscription whiteouts, and excludes OLM kinds from active resources.
  * Added a test helper to assert exact counts of generated whiteout resource files.
* **Bug Fixes**
  * Cleanup now checks and logs failures when removing temporary test artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/migtools/crane/pull/362)

### DIFF
--- a/e2e-tests/testdata/olm-multi-operator-resources.yaml
+++ b/e2e-tests/testdata/olm-multi-operator-resources.yaml
@@ -1,0 +1,38 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: olm-multi-og
+spec:
+  targetNamespaces: []
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: olm-multi-catalog
+spec:
+  sourceType: grpc
+  image: quay.io/operatorhubio/catalog:latest
+  displayName: Community Operators
+  publisher: OperatorHub.io
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: olm-multi-sub-certmgr
+spec:
+  channel: stable
+  installPlanApproval: Automatic
+  name: cert-manager
+  source: olm-multi-catalog
+  sourceNamespace: __NAMESPACE__
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: olm-multi-sub-cockroachdb
+spec:
+  channel: stable-v6.x
+  installPlanApproval: Automatic
+  name: cockroachdb
+  source: olm-multi-catalog
+  sourceNamespace: __NAMESPACE__
--- a/e2e-tests/tests/olm_whiteout_test.go
+++ b/e2e-tests/tests/olm_whiteout_test.go
@@ -276,7 +276,7 @@ var _ = Describe("OLM whiteout", func() {
 			Expect(utils.AssertNoKindsInOutput(paths.OutputDir, olmWhiteoutKinds)).NotTo(HaveOccurred())
 
 			By("Verify transform stage has whiteout resource files for both Subscriptions")
-			Expect(utils.AssertWhiteoutResourceFilesExist(paths.TransformDir, []string{"Subscription"})).NotTo(HaveOccurred())
+			Expect(utils.AssertWhiteoutResourceFilesExist(paths.TransformDir, []string{"Subscription", "CatalogSource", "OperatorGroup"})).NotTo(HaveOccurred())
 
 			By("Verify no partial whiteout: all OLM kinds excluded from active kustomize resources")
 			Expect(utils.AssertKindsNotInActiveKustomizeResources(paths.TransformDir, olmWhiteoutKinds)).NotTo(HaveOccurred())
--- a/e2e-tests/tests/olm_whiteout_test.go
+++ b/e2e-tests/tests/olm_whiteout_test.go
@@ -145,7 +145,9 @@ var _ = Describe("OLM whiteout", func() {
 					log.Printf("cleanup: failed to delete namespace: %v", err)
 				}
 				if paths.TempDir != "" {
-					os.RemoveAll(paths.TempDir)
+					if err := os.RemoveAll(paths.TempDir); err != nil {
+						log.Printf("cleanup: failed to remove temp dir %s: %v", paths.TempDir, err)
+					}
 				}
 			})
 
@@ -237,7 +239,9 @@ var _ = Describe("OLM whiteout", func() {
 					log.Printf("cleanup: failed to delete namespace: %v", err)
 				}
 				if paths.TempDir != "" {
-					os.RemoveAll(paths.TempDir)
+					if err := os.RemoveAll(paths.TempDir); err != nil {
+						log.Printf("cleanup: failed to remove temp dir %s: %v", paths.TempDir, err)
+					}
 				}
 			})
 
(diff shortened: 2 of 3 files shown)


e2e-tests/testdata/olm-multi-operator-resources.yaml
```diff
@@ -0,0 +1,38 @@
...
+spec:
+  targetNamespaces: []
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: olm-multi-catalog
+spec:
+  sourceType: grpc
+  image: quay.io/operatorhubio/catalog:latest
```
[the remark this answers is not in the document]
stillalearner: latest is fine here

e2e-tests/testdata/olm-multi-operator-resources.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: olm-multi-og
+spec:
+  targetNamespaces: []
```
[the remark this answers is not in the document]
stillalearner: intentional for multi namespaces.

e2e-tests/tests/olm_whiteout_test.go
```diff
@@ -194,6 +194,95 @@ var _ = Describe("OLM whiteout", func() {
...
+					if _, err := kubectlSrc.Run("delete", res.kind, res.name, "-n", namespace, "--ignore-not-found=true"); err != nil {
+						log.Printf("cleanup: failed to delete %s/%s: %v", res.kind, res.name, err)
+					}
+				}
+				By("Cleanup namespace and temp dir")
+				if _, err := kubectlSrc.Run("delete", "namespace", namespace, "--ignore-not-found=true"); err != nil {
+					log.Printf("cleanup: failed to delete namespace: %v", err)
+				}
+				if paths.TempDir != "" {
+					os.RemoveAll(paths.TempDir)
```
msajidmansoori12: @stillalearner can you please handle/log os.RemoveAll(paths.TempDir) errors ?
Right now cleanup failures are silent, which makes CI/debug triage harder. A small if err := os.RemoveAll(...); err != nil { log.Printf(...) } would be enough.

e2e-tests/tests/olm_whiteout_test.go
msajidmansoori12: @stillalearner can you please handle/log os.RemoveAll(paths.TempDir) errors ?
Right now cleanup failures are silent, which makes CI/debug triage harder. A small if err := os.RemoveAll(...); err != nil { log.Printf(...) } would be enough.

msajidmansoori12 (changes requested):

e2e-tests/tests/olm_whiteout_test.go
```diff
@@ -194,6 +194,95 @@ var _ = Describe("OLM whiteout", func() {
...
+			olmSpec = strings.ReplaceAll(olmSpec, "__NAMESPACE__", namespace)
+			Expect(kubectlSrc.ApplyYAMLSpec(olmSpec, namespace)).NotTo(HaveOccurred())
+
+			By("Wait for both operators to produce CSVs")
+			Eventually(func(g Gomega) {
+				out, err := kubectlSrc.Run("get", "csv", "-n", namespace, "-o", "jsonpath={.items[*].metadata.name}")
+				g.Expect(err).NotTo(HaveOccurred())
+				names := strings.Fields(out)
+				g.Expect(len(names)).To(BeNumerically(">=", 2), "expected at least 2 CSVs, got: %v", names)
+			}, "15m", "20s").Should(Succeed())
```
msajidmansoori12: @stillalearner 
len(CSVs) >= 2 is a good smoke signal, but it doesn’t prove both target subscriptions actually succeeded.
In OLM, one subscription can produce multiple CSV objects over time (current + replaced), so this check can pass even if one of the two subscriptions fails.

Could we add explicit checks for each subscription (olm-multi-sub-certmgr and olm-multi-sub-cockroachdb) — for example, assert both exist and each has status.currentCSV populated before running the crane pipeline?

This makes the test match its intent (“multiple operators”) and prevents false green runs where only one operator is truly installed.

msajidmansoori12 (changes requested):

e2e-tests/tests/olm_whiteout_test.go
```diff
@@ -194,6 +194,95 @@ var _ = Describe("OLM whiteout", func() {
...
+
+			By("Run crane export/transform/apply pipeline")
+			log.Printf("Running crane pipeline for namespace %s\n", namespace)
+			Expect(RunCranePipelineWithChecks(runner, namespace, paths)).NotTo(HaveOccurred())
+
+			By("Verify no OLM whiteout kinds in output")
+			Expect(utils.AssertNoKindsInOutput(paths.OutputDir, olmWhiteoutKinds)).NotTo(HaveOccurred())
+
+			By("Verify transform stage has whiteout resource files for both Subscriptions")
+			Expect(utils.AssertWhiteoutResourceFilesExist(paths.TransformDir, []string{"Subscription", "CatalogSource", "OperatorGroup"})).NotTo(HaveOccurred())
```
midays, line 288: IIUC, `AssertWhiteoutResourceFilesExist` returns success as soon as one file matches `Subscription_*`, so this test passes even if only one of the two Subscriptions gets a whiteout file which defeats the point of a multi-operator test, no?

midays, line 288: here the transform-dir check only lists `Subscription`, `CatalogSource`, `OperatorGroup`, while the output-dir check above uses the full `olmWhiteoutKinds` which includes `InstallPlan`, `ClusterServiceVersion`, `OperatorCondition`, Is that intentional?

e2e-tests/tests/olm_whiteout_test.go
```diff
@@ -194,6 +194,95 @@ var _ = Describe("OLM whiteout", func() {
...
+				names := strings.Fields(out)
+				g.Expect(len(names)).To(BeNumerically(">=", 2), "expected at least 2 CSVs, got: %v", names)
+			}, "15m", "20s").Should(Succeed())
+
+			By("Wait for InstallPlan to be created (OLM may bundle multiple operators into one)")
+			Eventually(func(g Gomega) {
+				out, err := kubectlSrc.Run("get", "installplan", "-n", namespace)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(strings.TrimSpace(out)).NotTo(ContainSubstring("No resources found"))
+			}, "15m", "20s").Should(Succeed())
```
midays, line 275: This passes as soon as any InstallPlan row exists, regardless of phase, so a Failed InstallPlan still lets the test proceed

perhaps a change like this is more reliable:
```suggestion
  Eventually(func(g Gomega) {
      out, err := kubectlSrc.Run("get", "installplan", "-n", namespace,
          "-o", "jsonpath={.items[*].status.phase}")
      g.Expect(err).NotTo(HaveOccurred())
      phases := strings.Fields(out)
      g.Expect(phases).NotTo(BeEmpty(), "no InstallPlans found")
      for _, p := range phases {
          g.Expect(p).To(Equal("Complete"), "InstallPlan in phase %q", p)
      }
  }, "10m", "20s").Should(Succeed())
```

midays (changes requested):
LGTM, left a few comments

e2e-tests/tests/olm_whiteout_test.go
stillalearner: Intentional yes, transform only has the 3 deployed kinds, output checks all OLM kinds as the other ones get created automatically afterwards, so its just a matter of timing

stillalearner: fixed this one.

midays (approved):
LGTM

Issue #189, opened 2026-03-24
Update transform&apply for cluster-level resources


aufi:
## Context

  Follow-up to #186. Depends on #188 and #256 being completed first.

  Cluster-scoped resources (ClusterRole, ClusterRoleBinding, CRDs, etc.)
  exported by `crane export` are currently not fully supported through the
  `transform` and `apply` stages. This issue tracks making them first-class
  citizens in the full pipeline.

  ## Goal

  Ensure cluster-scoped resources pass through `export → transform → apply`
  correctly: handled by transform plugins, rendered by kustomize, and written
  to a deterministic output layout that separates them from namespaced resources.

  ## Functional requirements

  1. Cluster-scoped resources from export must be consumed by transform — not
     skipped, misclassified, or assigned a fake namespace like `default`
  2. Transform plugins (especially `KubernetesPlugin`) must handle resources
     with empty `metadata.namespace` correctly
  3. Generated patch targets for cluster-scoped resources must omit `namespace`
  4. `crane apply` must place cluster-scoped resources under
     `output/resources/_cluster/` and namespaced resources under
     `output/resources/<namespace>/`
  5. `kubectl kustomize` must succeed on every produced stage with mixed
     namespaced + cluster-scoped inputs
  6. Whiteout and patch filtering logic must work for cluster-scoped resources
  7. Filenames and paths must be deterministic across reruns
  8. Recursive `kubectl apply -R -f output/` must work for the produced layout
  9. `crane apply` must support a `--skip-cluster-scoped` flag that excludes
     cluster-scoped resources from the rendered output (for non-admin migration
     scenarios where the user cannot apply cluster-level objects)

  ## Acceptance criteria

  - [ ] ClusterRole, ClusterRoleBinding, and CRD resources survive
        transform and appear in `output/resources/_cluster/`
  - [ ] Namespaced resources remain under `output/resources/<namespace>/`
  - [ ] `--skip-cluster-scoped` flag omits cluster-scoped resources from output
  - [ ] Patch target generation for cluster-scoped resources omits namespace
  - [ ] Recursive apply of output works for mixed namespaced + cluster-scoped cases
  - [ ] Unit tests: transform writer with cluster-scoped resources
  - [ ] Unit tests: apply split-output layout for cluster-scoped resources
  - [ ] Unit/integration test: mixed namespaced + cluster-scoped pipeline
  - [ ] E2E test: at least one scenario proving cluster-scoped dependency
        migration works end to end

PR #386, opened 2026-05-19
Add --skip-cluster-scoped flag to apply

stillalearner:
Phase 2 of [issue #189](https://github.com/migtools/crane/issues/189). Adds a `--skip-cluster-scoped` flag that excludes cluster-scoped resources from apply output, enabling non-admin migration scenarios where the user
   cannot apply cluster-level objects.

  ## Approach

  Post-build filtering — `kubectl kustomize` renders all resources (namespaced + cluster-scoped), then resources with empty `metadata.namespace` are filtered out before writing `output.yaml` and splitting into individual
   files. This is safer than pre-build filtering (modifying kustomization.yaml before kustomize runs) because it avoids any risk of breaking kustomize when patches reference cluster-scoped resources.

  ## Manual Testing on Minikube (all PASS)

  Setup: namespace `crane-apply-test` with nginx Deployment + ServiceAccount + ClusterRole + ClusterRoleBinding on src cluster.

  | Test | Result |
  |------|--------|
  | `crane apply` (default) — output contains `_cluster/` with ClusterRole + ClusterRoleBinding | PASS |
  | `crane apply --skip-cluster-scoped` — no `_cluster/` directory, only namespaced resources | PASS |
  | `output.yaml` with skip flag contains only ServiceAccount + Deployment (no ClusterRole) | PASS |
  | `kubectl apply -R -f output/resources/ --dry-run=client` works for both outputs | PASS |
  | Log messages: "Skipping cluster-scoped resource ClusterRole/..." printed for each filtered resource | PASS |

  ## Unit Tests Added (4 tests in `internal/apply/kustomize_test.go`)

  | Test | Verifies |
  |------|----------|
  | `TestSplitMultiDocYAML_SkipClusterScoped` | Mixed resources + skip flag: `_cluster/` not created, namespaced resources preserved, filtered YAML has no ClusterRole |
  | `TestSplitMultiDocYAML_ClusterScopedIncluded` | Same input without skip flag: both `_cluster/` and namespace directories created |
  | `TestSplitMultiDocYAML_OnlyClusterScoped` | All cluster-scoped + skip flag: empty output, no `resources/` directory |
  | `TestFilterClusterScopedResources_PreservesNamespaced` | Deployment + ServiceAccount kept, ClusterRole removed |

## Summary by CodeRabbit

* **New Features**
  * Added a --skip-cluster-scoped option to the apply command to optionally exclude cluster-scoped Kubernetes resources from generated outputs; when enabled, only namespaced resources are emitted and cluster-scoped kinds are filtered out.
* **Bug Fixes**
  * Improved splitting of multi-document YAML to correctly handle empty/filtered outputs and surface encoding/close errors.
* **Tests**
  * Added tests covering mixed, only-cluster-scoped, and only-namespaced scenarios to verify filtering and output layout.

[Review Change Stack](https://app.coderabbit.ai/change-stack/migtools/crane/pull/386?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/apply/kustomize.go
+++ b/internal/apply/kustomize.go
@@ -163,7 +163,9 @@ func (k *KustomizeApplier) filterClusterScopedResources(yamlData []byte) ([]byte
 		if err := encoder.Encode(doc); err != nil {
 			return nil, fmt.Errorf("failed to encode YAML document: %w", err)
 		}
-		encoder.Close()
+		if err := encoder.Close(); err != nil {
+			return nil, fmt.Errorf("failed to close YAML encoder: %w", err)
+		}
 		docBytes := buf.Bytes()
 
 		jsonData, err := yaml.YAMLToJSON(docBytes)
--- a/internal/apply/kustomize.go
+++ b/internal/apply/kustomize.go
@@ -219,7 +219,9 @@ func (k *KustomizeApplier) splitMultiDocYAMLToFiles(yamlData []byte) error {
 		if err := encoder.Encode(doc); err != nil {
 			return fmt.Errorf("failed to encode YAML document: %w", err)
 		}
-		encoder.Close() // Flush the encoder
+		if err := encoder.Close(); err != nil {
+			return fmt.Errorf("failed to close YAML encoder: %w", err)
+		}
 		docBytes := buf.Bytes()
 
 		// Convert YAML to JSON to extract metadata
--- a/internal/apply/kustomize.go
+++ b/internal/apply/kustomize.go
@@ -53,6 +53,14 @@ func (k *KustomizeApplier) ApplySingleStage(stageName string) error {
 		return fmt.Errorf("kustomize build failed for stage %s: %w", stageName, err)
 	}
 
+	// Filter cluster-scoped resources if requested
+	if k.SkipClusterScoped {
+		output, err = k.filterClusterScopedResources(output)
+		if err != nil {
+			return fmt.Errorf("failed to filter cluster-scoped resources: %w", err)
+		}
+	}
+
 	// Write output to output directory
 	outputPath := filepath.Join(k.OutputDir, stageName+".yaml")
 	if err := os.MkdirAll(k.OutputDir, 0700); err != nil {
--- a/internal/apply/kustomize.go
+++ b/internal/apply/kustomize.go
@@ -19,10 +19,11 @@ import (
 
 // KustomizeApplier applies transformations using embedded kustomize
 type KustomizeApplier struct {
-	Log           *logrus.Logger
-	TransformDir  string
-	OutputDir     string
-	KustomizeArgs []string
+	Log               *logrus.Logger
+	TransformDir      string
+	OutputDir         string
+	KustomizeArgs     []string
+	SkipClusterScoped bool
 }
 
 // ApplySingleStage applies a single transform stage to produce output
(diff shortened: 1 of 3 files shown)


internal/apply/kustomize.go
```diff
@@ -159,6 +168,59 @@ func ValidateKubectlAvailable() error {
...
+		}
+
+		u := unstructured.Unstructured{}
+		if err := u.UnmarshalJSON(jsonData); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal resource: %w", err)
+		}
+
+		if u.GetNamespace() == "" {
+			k.Log.Infof("Skipping cluster-scoped resource %s/%s (--skip-cluster-scoped)", u.GetKind(), u.GetName())
+			continue
```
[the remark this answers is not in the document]
stillalearner: The filter doesn't operate on arbitrary YAML files. It operates on the output of kubectl kustomize, which is fed resources that came from crane export. Export reads resources from the Kubernetes API server, where every
   namespaced resource already has metadata.namespace populated — the API server always sets it. That value is preserved through export → transform → kustomize build.

  There's no point in the pipeline where a namespaced resource would lose its namespace:

  1. Export — reads from API server, namespace is always set
  2. Transform — plugins operate via JSONPatch on existing fields, they don't remove namespace
  3. Kustomize build — preserves namespace from source YAML

internal/apply/kustomize.go
msajidmansoori12, line 55: @stillalearner perhaps , this logic ```if k.SkipClusterScoped {
		output, err = k.filterClusterScopedResources(output)
		if err != nil {
			return fmt.Errorf("failed to filter cluster-scoped resources: %w", err)
		}
	}```
should be added in this function as well for consistency , in case it is used later on and missed.

msajidmansoori12 (changes requested):

internal/apply/kustomize.go
stillalearner: sure, let me add, 

but we can consider removing this function as its kind of dead code..

msajidmansoori12: @stillalearner agreed, if it's dead , we can maybe remove it in a follow up PR.

internal/apply/kustomize.go
```diff
@@ -159,6 +176,59 @@ func ValidateKubectlAvailable() error {
...
+		}
+
+		var buf bytes.Buffer
+		encoder := yamlv3.NewEncoder(&buf)
+		encoder.SetIndent(2)
+		if err := encoder.Encode(doc); err != nil {
+			return nil, fmt.Errorf("failed to encode YAML document: %w", err)
+		}
+		encoder.Close()
```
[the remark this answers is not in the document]
msajidmansoori12: @stillalearner can you please address this, if it makes sense. Thanks.

msajidmansoori12 (approved):

internal/apply/kustomize.go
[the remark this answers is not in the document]
stillalearner: fixed

internal/apply/kustomize.go
```diff
@@ -121,6 +138,61 @@ func (k *KustomizeApplier) runKustomizeBuild(dir string) ([]byte, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert YAML to JSON: %w", err)
+		}
+
+		u := unstructured.Unstructured{}
+		if err := u.UnmarshalJSON(jsonData); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal resource: %w", err)
+		}
+
+		if u.GetNamespace() == "" {
```
aufi, line 181: Overall PR looks good, we might need consider/challenge this way of selecting cluster scoped resources (in apply step).

There is a valid use case with gitops to have output  resources without specific namespace (so it can be deloyed to the one chosen in gitops tool (IIRC namespace in resources manifests take priority in e.g. argocd) over the one chosen during the deployment).

But I'm not sure if there is other reasonable way identify cluster level resources.

stillalearner: Good point about the GitOps use case. Today I have gone with this filter as it is safe because crane's pipeline guarantees metadata.namespace is always populated for namespaced resources, as export reads from the API server (which always serializes namespace), transform preserves it, and kustomize preserves it. So namespace == "" reliably identifies cluster-scoped resources throughout the pipeline. 

If we ever add a feature to strip namespaces from output for GitOps compatibility, we'd need to switch to Kind-based detection at that point, either using a known cluster-scoped kinds list
  (ClusterRole, ClusterRoleBinding, CRD, Namespace, etc.) or API discovery.

aufi (approved):
LGTM

Issue #370, opened 2026-05-12
Update transform with instruction file and multiple plugins


aufi:
Based on current approach creating a single transformation stage per `crane transform` call and conversation on https://github.com/konveyor/enhancements/pull/288, it is needed update CLI to be more user friendly with building transform stages pipeline. It is expected do a design work first to find the right way how to deliver this.

## Expected results
It is possible create more stages with a single `crane transform` command call while keeping it clear and easy to use.
It is possible provide a transformation YAML file with `stages` key containing list of stages/plugins that should be used.
Both need to be user friendly and documented.

## Resources
https://github.com/konveyor/enhancements/pull/288
https://github.com/aufi/crane/blob/b53e2565959c54f9f677f2b622c433fce4f0a336/docs/transform-scenarios/03-multistage.md
https://github.com/aufi/crane/tree/transform-instructions-file/instructions_file_notes

aufi:
For the Task 3 (`Add an instructions / migration descriptor file as input for transform command`), there is a draft doc on how it could look like: https://github.com/aufi/crane/tree/transform-instructions-file/instructions_file_notes

Please take it as an example, not exact&confirmed requirements. We need implement something simple and useful, so if the draft describes what it should do, but e.g. if there is a need to have 
```
stages:
    - name: "10_KubernetesPlugin"
      type: "plugin"
```

or just (order is implicit in the array, Plugin suffix for the stage type too)

```
stages:
    - name: "KubernetesPlugin"
```
 or even (if any additional args for plugin stages will never be needed)
```
stages:
    - KubernetesPlugin
    - SomeOtherPlugin
    - CustomModification
```

is up to discussion&experiments, I'm happy to discuss it.

msajidmansoori12:
[Doc](https://docs.google.com/document/d/1DNZVO2jIgoKJc3MSM0mzf-tQCL6qmmsCFJIPOs1kg8U/edit?tab=t.0) with current implementation of transform descriptor file and some e2e migration tests.

Issue #331, opened 2026-05-03
[BUG] OpenShiftPlugin not included in default stage creation when installed


midays:
**What version of crane are you running, and what are your clutsters+platform**
crane: main branch
OpenShiftPlugin: v0.0.4
Platform: macOS arm64

**What did you expect to happen?**
When OpenShiftPlugin is installed in the plugin directory and crane transform
is run without explicit --stage flags, crane should automatically create and
run stages for all discovered plugins, e.g.:
  10_KubernetesPlugin
  20_OpenShiftPlugin

**What actually happened?**
Despite OpenShiftPlugin being installed and correctly discovered
(debug logs show "Loaded 2 plugin(s)"), crane only auto-creates and runs
the 10_KubernetesPlugin stage:
```
  INFO No existing stages found, creating default stage: 10_KubernetesPlugin
  INFO Executing stage 1/1: 10_KubernetesPlugin
  INFO Successfully completed 1 stage(s)
```
The OpenShiftPlugin is silently skipped during default stage creation.
The only workaround is to manually invoke a second transform with an
explicit stage name:

```
  crane transform \
    --export-dir ./export/resources/<ns> \
    --plugin-dir ~/.local/share/crane/plugins \
    --stage 20_OpenShiftPlugin
```

**Reproduction steps:**
1) Install OpenShiftPlugin: `$crane plugin-manager add OpenShiftPlugin`
2) Verify both plugins are discovered: `$crane transform list-plugins`
   (shows KubernetesPlugin and OpenShiftPlugin)
3) Run: `$crane transform --export-dir ./export/resources/<ns> --plugin-dir ~/.local/share/crane/plugins`
4) Observe: only 10_KubernetesPlugin stage is created and executed
5) OpenShiftPlugin is never run unless --stage is explicitly specified

PR #395, opened 2026-05-21
Transform stages run all by default or positional args

aufi:
Default execution of `crane transform` creates stages for all available plugins (instead of just default KubernetePlugin). This could be optionally restricted with --skip-plugins arg.

Also allowing create multiple stages per single `crane transform` command call and its update to positional args.  **Important:** breaking change from previous `--stage`, docs update will follow in another PR.

Related to https://github.com/migtools/crane/issues/370 (Task 1 https://github.com/migtools/crane/issues/409 and Task 2 https://github.com/migtools/crane/issues/410)

## Summary by CodeRabbit

* **New Features**
  * Transform and Apply now accept positional stage arguments (e.g., command [stage...]) and support selecting multiple stages; when no stages exist, transforms auto-create one default stage per discovered plugin and run them.

* **Bug Fixes**
  * Enforced mutual exclusivity between positional stage args and instruction-file input. Generated stage names/paths are validated to prevent invalid names and path traversal; invalid plugins are skipped and created stages are tracked.

* **Tests**
  * Added/updated tests for conflicts, path traversal protection, name validation, stage creation, ordering, and multi-stage selection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/migtools/crane/pull/395?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


aufi:
Got conflicts after https://github.com/migtools/crane/pull/412 merge, so rather added also remaning part of executing multiple stages by a single command call into this PR. Up for review.

cmd/transform/transform.go
```diff
@@ -407,3 +402,233 @@ func (o *Options) ensureStagesHaveOutput(orchestrator *internalTransform.Orchest
...
+		}
+
+		// Not found - try to create new stage for this plugin
+		log.Infof("Stage for %q not found, attempting to create", requested)
+
+		// Load plugins to verify it exists
+		allPlugins, err := plugin.GetFilteredPlugins(pluginDir, o.SkipPlugins, log)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load plugins: %w", err)
+		}
```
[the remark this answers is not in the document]
aufi: Skipping this minor issue for now.

cmd/transform/transform.go
```diff
@@ -407,3 +409,267 @@ func (o *Options) ensureStagesHaveOutput(orchestrator *internalTransform.Orchest
...
+) ([]string, error) {
+	var stageNames []string
+
+	// Sort plugins by name for deterministic ordering
+	sort.Slice(allPlugins, func(i, j int) bool {
+		return allPlugins[i].Metadata().Name < allPlugins[j].Metadata().Name
+	})
+
+	// Assign priority to each plugin
+	// Start at 10, increment by 5 for each plugin
```
aufi, line 428: Default (all) plugins stages get order increase by 5, I wasn't sure how many there will be so used smaller number (but no strong opinion about it).

cmd/apply/apply.go
```diff
@@ -140,12 +142,14 @@ func (o *Options) run() error {
...
+	if len(o.RequestedStages) > 0 {
+		// User specified specific stages via positional arguments
+		// For apply, we use FilterStages to match by directory or plugin name
+		// (we don't create new stages in apply, only in transform)
 		selector = internalTransform.StageSelector{
-			Stage: o.Flags.Stage,
+			Stages: o.RequestedStages,
 		}
-		log.Infof("Applying stage: %s", o.Flags.Stage)
+		log.Infof("Applying %d stage(s): %v", len(o.RequestedStages), o.RequestedStages)
```
msajidmansoori12, line 195: @aufi Hi , Here we pass requested stage args directly to apply.
Could we validate that all requested stages resolve first?
Right now a typo can be silently ignored if at least one valid stage matches, and apply still succeeds which is risky because user may think both were applied.

Could we validate all requested stages first and return a clear error listing unknown args?

Eg: 
```
go run main.go apply KubernetesPlugin TypoStage
INFO[0000] Applying 2 stage(s): [KubernetesPlugin TypoStage] 
INFO[0000] Applying final stage: 10_KubernetesPlugin    
INFO[0000] Successfully applied final stage to /Users/mmansoor/Desktop/git_repos/crane-testing/crane/output/output.yaml 
```

msajidmansoori12 (changes requested):

cmd/apply/apply.go
aufi: Right, thanks, updated.

msajidmansoori12 (approved):
/lgtm thanks for incorporating the changes @aufi

stillalearner:
hi @aufi , can you please mention in description that there is a breaking change here
--stage flag removed

cmd/transform/transform.go
stillalearner: @aufi i think we can keep it increment by 10, for consistency..

cmd/transform/transform.go
```diff
@@ -407,3 +408,348 @@ func (o *Options) ensureStagesHaveOutput(orchestrator *internalTransform.Orchest
...
+	for _, stage := range existingStages {
+		if stage.DirName == name {
+			return stage, true
+		}
+	}
+	return internalTransform.Stage{}, false
+}
+
+// findStagesByPluginName returns all stages matching a plugin name
+func findStagesByPluginName(existingStages []internalTransform.Stage, pluginName string) []internalTransform.Stage {
```
stillalearner: findStagesByPluginName & findStagesByBaseName are duplicate functions.

can we remove one of those

aufi: Good point, updated.

stillalearner (approved):
LGTM

aufi:
This PR fixes also https://github.com/migtools/crane/issues/331

Issue #369, opened 2026-05-12
Embed kustomize into crane


aufi:
Crane `transform` and `apply` subcommands use kustomize tool (via `kubectl kustomize`) to perform transformations of k8s manifests. Based on conversation on https://github.com/konveyor/enhancements/pull/288, we want change current approach from calling external binary to embed kustomize as a golang dependency and call it internally to create and apply transformations. Pay attention to flexible kustomize custom args setup to allow kustomize components and helm templates like allowed in https://github.com/migtools/crane/pull/358.

## Tasks
1. Embed kustomize into crane binary and ensure there are no external dependencies on `kubectl` or `oc`.
2. Update `crane version` to provide also kustomize library version.
3. Provide evaluation on crane binary size change and added go dependencies assessment (regarding to its maintenance and updates in future).
4. Finally update the enhancement with kustomize embeded instead of called externally.

## Expected result
There is no `kubectl` or `oc` binary execution in the crane sources and E2E test pass without `kubectl` binary installed in the system.

## Resources
https://github.com/konveyor/enhancements/pull/288#pullrequestreview-4263772946
https://github.com/aufi/move-crane/blob/main/notes/kubectl-call-or-embed.md

stillalearner:
kustomize/api was already an indirect dependency so embedding it didn't change binary size.

stillalearner:
Enhancement doc updates have been added as suggestions to the enhancements proposal PR.

PR #413, opened 2026-05-26
Added kustomize version

stillalearner:
# PR: Add kustomize version to `crane version` output

  Task 2 of [issue #369](https://github.com/migtools/crane/issues/369). Now that kustomize is embedded, users need visibility into which kustomize version is baked into the crane binary.

  ## Output

  $ crane version
  crane:
      Version: v0.0.6
  crane-lib:
      Version: v0.0.10
  kustomize:
      Version: v5.8.1

  ## Binary Size Impact

  - Before: 81,710,386 bytes (78M)
  - After: 81,728,770 bytes (78M)
  - Increase: ~18KB (0.02%)

  ## Files Changed

  | File | Lines | Change |
  |------|-------|--------|
  | `internal/buildinfo/buildinfo.go` | +16 | `KustomizeVersion` via CLI module + `debug.ReadBuildInfo()` |
  | `internal/buildinfo/buildinfo_test.go` | +57 | 3 unit tests |
  | `cmd/version/version.go` | +3 | Print kustomize version, update Short description |
  | `cmd/version/version_test.go` | +46 | 1 test for version output |
  | `go.mod` / `go.sum` | +2 | Added kustomize CLI module dependency |

### DIFF
--- a/internal/buildinfo/buildinfo.go
+++ b/internal/buildinfo/buildinfo.go
@@ -4,6 +4,9 @@ import (
 	"runtime/debug"
 
 	cranelibversion "github.com/konveyor/crane-lib/version"
+	// Blank import to keep the kustomize CLI module in the dependency graph
+	// so debug.ReadBuildInfo() can report its version.
+	_ "sigs.k8s.io/kustomize/kustomize/v5/commands/version"
 )
 
 var (
@@ -22,7 +25,7 @@ func readKustomizeVersion() string {
 		return "unknown"
 	}
 	for _, dep := range info.Deps {
-		if dep.Path == "sigs.k8s.io/kustomize/api" {
+		if dep.Path == "sigs.k8s.io/kustomize/kustomize/v5" {
 			return dep.Version
 		}
 	}
--- a/internal/buildinfo/buildinfo.go
+++ b/internal/buildinfo/buildinfo.go
@@ -1,6 +1,8 @@
 package buildinfo
 
 import (
+	"runtime/debug"
+
 	cranelibversion "github.com/konveyor/crane-lib/version"
 )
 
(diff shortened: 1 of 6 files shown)


msajidmansoori12 (approved):
/lgtm

aufi:
It looks to me, this Version provides kustomize API version (~v0.20), instead of "own" kustomize (~v5.8), see https://github.com/kubernetes-sigs/kustomize/releases

@stillalearner please check if it is possible return the own kustomize version (or clarify that the 0.20-like version is more suitable), thanks!


```
$ crane version
crane:
	Version: v0.0.6
crane-lib:
	Version: v0.0.10
kustomize:
	Version: v0.20.1
```

stillalearner:
> It looks to me, this Version provides kustomize API version (~v0.20), instead of "own" kustomize (~v5.8), see https://github.com/kubernetes-sigs/kustomize/releases
> 
> @stillalearner please check if it is possible return the own kustomize version (or clarify that the 0.20-like version is more suitable), thanks!
> 
> ```
> $ crane version
> crane:
> 	Version: v0.0.6
> crane-lib:
> 	Version: v0.0.10
> kustomize:
> 	Version: v0.20.1
> ```


So, crane currently only imports the API module, and not the CLI , so we dont have a direct way to get the CLI version. 
Either we maintain a mapping, Or import the CLI module, just for this version, which also doesnt make sense. OR we just keep this API version and mention in the output that this is api version. 

kustomize (API version):
     Version: v0.20.1

internal/buildinfo/buildinfo.go
```diff
@@ -1,11 +1,33 @@
 package buildinfo
 
 import (
+	"runtime/debug"
+
 	cranelibversion "github.com/konveyor/crane-lib/version"
+	// Blank import to keep the kustomize CLI module in the dependency graph
+	// so debug.ReadBuildInfo() can report its version.
+	_ "sigs.k8s.io/kustomize/kustomize/v5/commands/version"
```
aufi, line 9: Thanks for commenting this import as it is not fully obvious how the kustomize import is used.

aufi (approved):
Thanks for update Sachin! Works well for me & crane binary size was not affected, LGTM.

Using kustomize dep import just for version is probably not an ideal way, but we needed get the version somewhere, so no problem with merging.

```
$ ./crane version
crane:
	Version: v0.0.6
crane-lib:
	Version: v0.0.10
kustomize:
	Version: v5.8.1
```

PR #443, opened 2026-06-02
fix(export): implement per-request timeout and fail-fast behavior acr…

Tamar-Dinavetsky:
### Description
Addresses the issue where `--request-timeout` was being ignored during the resource extraction phase due to hardcoded `context.Background()` calls.

Following the team's feedback, this implements the **per-request timeout** and strict **fail-fast behavior**:
- Replaced global context logic with a fresh context timeout instantiated for each individual API/resource request loop (`discover.go`, `export.go`, and `crd.go`).
- Updated error handling inside the listing loops to immediately abort execution and return the error on deadline/timeout exceedance, preventing partial or invalid directory exports.

### Testing
- Tested manually with various timeout windows (`100ms`, `50ms`, `10ms`).
- Confirmed that healthy requests pass within their individual windows, while slow requests drop immediately and fail fast, aborting the rest of the operation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fail-fast on request timeouts during export discovery and related-resource collection to avoid hanging operations.
  * More reliable detection and reporting of timeout/deadline-exceeded errors.

* **Improvements**
  * Export honors configured per-request timeouts across discovery, listing, and per-item retrieval flows.
  * CLI suppresses noisy usage output when operations fail due to timeouts.

* **Tests**
  * Existing tests updated and new timeout-focused tests added to verify fail-fast behavior and error reporting.


aufi (commented):
Thanks for the PR @Tamar-Dinavetsky! I'd suggest address coderabbit reports (primary major like updating command exit code on timeout).

cmd/export/crd.go
```diff
@@ -112,7 +113,14 @@ func collectRelatedCRDs(resources []*groupResource, dynamicClient dynamic.Interf
 	out := make([]*groupResource, 0, len(seen))
 	var outErrs []*groupResourceError
 	for crdName := range seen {
-		obj, err := crdClient.Get(context.Background(), crdName, metav1.GetOptions{})
+		// Create fresh context with timeout for each CRD Get request
+		ctx := context.Background()
+		if requestTimeout > 0 {
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, requestTimeout)
+			defer cancel()
```
midays: `defer cancel()` inside a for loop doesn't cancel after each iteration, all deferred calls from the loop stack up and run only when `collectRelatedCRDs` returns, leaking context resources for the duration of the loop. 
I'm not sure this is the best solution for this, but maybe replace with an explicit cancel() call after the Get after [line123](https://github.com/migtools/crane/pull/443/changes#diff-cc1dd541dc7ecfd7532bfaeac8f6fe0b3c5b5c4f73f4448c1c70c592157f1f1fR123)

cmd/export/discover.go
```diff
@@ -289,30 +295,45 @@ func isAdmittedResource(gv schema.GroupVersion, resource metav1.APIResource) boo
...
+func getObjects(requestTimeout time.Duration, g *groupResource, namespace string, labelSelector string, d dynamic.Interface, logger logrus.FieldLogger) (*unstructured.UnstructuredList, error) {
 	c := d.Resource(schema.GroupVersionResource{
 		Group:    g.APIGroup,
 		Version:  g.APIVersion,
 		Resource: g.APIResource.Name,
 	})
-	p := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+	p := pager.New(func(pagerCtx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		// Create a fresh context with timeout for each page request
+		ctx := context.Background()
```
midays:
```suggestion
		ctx := pagerCtx
```

is there a reason a fresh context? The per-page context should derive from pagerCtx rather than context.Background(). 
Replacing it with a fresh background context severs the cancellation chain, meaning if the outer `p.List` context is cancelled (e.g. user interrupt), individual page requests won't notice and will keep running.

midays (changes requested):
nice PR, left 2 comments, otherwise LGTM

cmd/export/discover.go
```diff
@@ -289,30 +295,45 @@ func isAdmittedResource(gv schema.GroupVersion, resource metav1.APIResource) boo
...
 		Resource: g.APIResource.Name,
 	})
-	p := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+	p := pager.New(func(pagerCtx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		// Create a fresh context with timeout for each page request
+		ctx := pagerCtx
+		if requestTimeout > 0 {
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, requestTimeout)
+			defer cancel()
```
aufi: The defer cancel() inside the pager callback function (line 304-309) does not execute after each page request but only when the entire pager completes, causing a memory/goroutine leak when paginating through multiple pages — fix by declaring var cancel context.CancelFunc outside
  the if block so it's always defined and safe to defer.

Tamar-Dinavetsky: Updated. The context leaks in both the discovery pager callback and the `iterateItemsByGet` loop have been resolved by ensuring proper cancellation immediately after the API calls. 

The PR is ready for another look.

aufi: Thanks, please resolve conflict, so we can re-review and approve.

Tamar-Dinavetsky: Done. Ready for review

midays (approved):
LGTM

Issue #427, opened 2026-06-01
[BUG] [validate] manifests scanned before kubeconfig/context is validated


midays:
## Description

When an invalid `--context` is passed, `crane validate` scans the input directory and prints progress output before surfacing the error. The user sees a successful-looking log line followed by a fatal error, which is confusing.

## Steps to Reproduce

```bash
./crane validate --context does-not-exist -i output/
```

## Actual Result

```
INFO[0000] Scanned 3 distinct GVK+namespace tuples
Error: creating discovery client: context "does-not-exist" does not exist
```

## Expected Result

The context/kubeconfig should be validated before scanning starts. Error should surface immediately with no prior progress output:

```
Error: creating discovery client: context "does-not-exist" does not exist
```

## Root Cause

`Complete()` only loads the raw kubeconfig config, it does not validate the context. The discovery client is created later in `Run()` after `ScanManifests()` has already executed. The fix would be to attempt discovery client creation in `Validate()` or early in `Complete()` before scanning begins.

## Environment

- 2 minikube clusters, namespace admin (`src-dev` / `tgt-dev`)

PR #445, opened 2026-06-03
Fixed bug 427: Manifests scanned before kubeconfig/context

stillalearner:
## The Problem

  ```bash
  $ crane validate --context does-not-exist -i output/
  INFO[0000] Scanned 3 distinct GVK+namespace tuples      ← misleading progress
  Error: creating discovery client: context "does-not-exist" does not exist
```

##  The Fix

  Replaced ToRawKubeConfigLoader().RawConfig() with ToDiscoveryClient() in Complete(). The old call only parsed the kubeconfig file without checking whether the specified context exists. The new call creates a discovery client
  for the specific context, which validates the context name before Run() starts scanning.

  $ crane validate --context does-not-exist -i output/
  Error: creating discovery client: context "does-not-exist" does not exist

No scan output before the error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Live validation now performs an earlier discovery client check and fails faster with clearer discovery/connection errors.
  * Offline validation (API-resources mode) behavior is unchanged.

* **Tests**
  * Added tests ensuring invalid cluster/context configurations fail early during validation and that offline/API-resources mode continues to skip live checks.

### DIFF
--- a/cmd/validate/validate.go
+++ b/cmd/validate/validate.go
@@ -29,15 +29,15 @@ type ValidateOptions struct {
 	genericclioptions.IOStreams
 }
 
-// Complete loads the kubeconfig to ensure the target cluster is reachable.
+// Complete loads the kubeconfig and validates the discovery client is reachable.
 // Skipped in offline mode (--api-resources).
 func (o *ValidateOptions) Complete(c *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


aufi (approved):
LGTM

issue: https://github.com/migtools/crane/issues/427

PR #450, opened 2026-06-03
Fix bug 431: validate directory overwrites without warning

### LINKED ISSUE #431: [BUG] [validate]  --validate-dir silently overwrites existing report on consecutive runs
## Description

When `--validate-dir` already contains a `report.json` from a previous run, it is silently overwritten with no warning. A user running `crane validate` multiple times could lose previous results without realizing it.

## Steps to Reproduce

```bash
# First run
./crane validate --context tgt-dev -i output/ --validate-dir validate-results/
# report.json written

# Second run -- overwrites silently
./crane validate --context tgt-dev -i output/ --validate-dir validate-results/
# No warning, previous report.json gone
```

## Actual Result

Existing `report.json` is overwritten silently. No warning is printed, no prompt, no indication that previous results were lost.

## Expected Result

One of:
1. Print a warning when overwriting an existing report: `WARN: overwriting existing report in <dir>`
2. Add a `--overwrite` flag that must be explicitly passed to allow overwriting
3. Use timestamped report filenames (e.g. `report-20260531-1821.json`) to preserve history

## Environment

- 2 minikube clusters, namespace admin (`src-dev` / `tgt-dev`)

stillalearner:
# PR: Fix bug 431: --validate-dir silently overwrites existing report on consecutive runs

  Fixes [#431](https://github.com/migtools/crane/issues/431). 
  Previous validation reports and failures were silently overwritten on consecutive runs, losing audit history.

  ## The Problem

  ```bash
  crane validate --context tgt -i output/ --validate-dir validate/
  # writes validate/report.json

  crane validate --context tgt -i output/ --validate-dir validate/
  # silently overwrites report.json — previous results lost
 ```
 
 ## The Fix

  Before writing new results, existing report.json and failures/ are renamed using the report's modification timestamp. Both are archived with the same timestamp derived from report.json to guarantee they stay grouped.

  First run:
  validate/
  ├── report.json
  └── failures/

  Second run:
  validate/
  ├── report.json                       ← latest
  ├── report-20260603-100942.json       ← previous (renamed)
  ├── failures/                         ← latest
  └── failures-20260603-100942/         ← previous (renamed, same timestamp)

  Third run:
  validate/
  ├── report.json                       ← latest
  ├── report-20260603-100942.json       ← first run
  ├── report-20260603-112015.json       ← second run
  ├── failures/                         ← latest
  ├── failures-20260603-100942/         ← first run
  └── failures-20260603-112015/         ← second run

  - report.json always exists at a stable path for scripts
  - Previous results preserved with timestamps for audit
  - Timestamp is the report file's modification time (when it was originally written), not the current time
  - Report and failures share the same timestamp so they stay paired
  - No data duplication on disk

  Manual Testing on Minikube (all PASS)

## Summary by CodeRabbit

* **New Features**
  * Previous validation reports and failure directories are now automatically archived before new reports are generated. Archived items use a shared timestamp-based naming to preserve validation run history and avoid overwriting.

* **Tests**
  * Added unit tests covering timestamp computation and archiving behavior, including cases for files, directories, format switches, and no-previous-results.

### DIFF
--- a/cmd/validate/validate.go
+++ b/cmd/validate/validate.go
@@ -123,18 +123,10 @@ func (o *ValidateOptions) Run() error {
 	reportPath := filepath.Join(o.validateDir, "report."+reportExt)
 	failuresDir := filepath.Join(o.validateDir, "failures")
 
-	// Archive previous results using report's timestamp so report and failures
-	// are grouped by the same timestamp even if they were written at slightly
-	// different times within the same run.
-	if ts, err := getArchiveTimestamp(reportPath); err != nil {
-		return fmt.Errorf("checking previous report: %w", err)
-	} else if ts != "" {
-		if err := archiveWithTimestamp(reportPath, ts, log); err != nil {
-			return fmt.Errorf("archiving previous report: %w", err)
-		}
-		if err := archiveWithTimestamp(failuresDir, ts, log); err != nil {
-			return fmt.Errorf("archiving previous failures: %w", err)
-		}
+	// Archive previous results. Check for report in any format (json or yaml)
+	// to handle the case where the user switches --output format between runs.
+	if err := archivePreviousResults(o.validateDir, failuresDir, log); err != nil {
+		return err
 	}
 
 	reportFile, err := os.Create(reportPath)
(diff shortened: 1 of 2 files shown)


nachandr:
Since the validate directory will accumulate many archived results, could we have an archive retention policy (e.g., keep last N archives, or archives from last M days) ?

stillalearner:
> Since the validate directory will accumulate many archived results, could we have an archive retention policy (e.g., keep last N archives, or archives from last M days) ?

We are just renaming the previous report and directory, 

I think it is the headache of user of how many reports he wants to keep in his local. 

We are not keeping anything.

cmd/validate/validate.go
```diff
@@ -151,6 +167,45 @@ func (o *ValidateOptions) Run() error {
...
+// as a timestamp string. Returns empty string if the path does not exist.
+func getArchiveTimestamp(path string) (string, error) {
+	info, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return "", nil
+	}
+	if err != nil {
+		return "", fmt.Errorf("checking %q: %w", path, err)
+	}
+	return info.ModTime().Format("20060102-150405"), nil
```
aufi, line 172: 👍 nice using the file modtime

cmd/validate/validate.go
```diff
@@ -118,8 +118,25 @@ func (o *ValidateOptions) Run() error {
...
 	reportExt := o.outputFormat
 	reportPath := filepath.Join(o.validateDir, "report."+reportExt)
+	failuresDir := filepath.Join(o.validateDir, "failures")
+
+	// Archive previous results using report's timestamp so report and failures
+	// are grouped by the same timestamp even if they were written at slightly
+	// different times within the same run.
+	if ts, err := getArchiveTimestamp(reportPath); err != nil {
+		return fmt.Errorf("checking previous report: %w", err)
+	} else if ts != "" {
```
aufi: Not a merge blocker, just maybe a nit for better readable condition.

Return from condition "true" block and then `else if`, could be something like following more clear:
```
ts, err := getArchiveTimestamp(reportPath)
if err != nil {
    return fmt.Errorf("checking previous report: %w", err)
} 

// something to archive 
if ts != "" {
    if err := archiveWithTimestamp(reportPath, ts, log); err != nil {
        return fmt.Errorf("archiving previous report: %w", err)
    }
    if err := archiveWithTimestamp(failuresDir, ts, log); err != nil {
        return fmt.Errorf("archiving previous failures: %w", err)
    }
}
```

Not requiring the change, current code looks to work, so it is up to your consideration.

aufi (approved):

midays (approved):
no issues, solid test case

cmd/validate/validate.go
stillalearner: Rule 1: DONT TOUCH THE WORKING CODE

PR #458, opened 2026-06-05
Inject proper crane version from tag in CI

msajidmansoori12:
## Summary by CodeRabbit

* **Chores**
  * Release builds now write compiled artifacts to a dedicated distribution directory.
  * Build process embeds a computed build version (tag or commit-based) into artifacts and validates its presence.
  * CI will fail if the expected version metadata is not found in the produced binary.

* **New Features**
  * CLI version output now includes the build commit SHA alongside the version for clearer traceability.

### DIFF
--- a/internal/buildinfo/buildinfo.go
+++ b/internal/buildinfo/buildinfo.go
@@ -10,9 +10,12 @@ import (
 )
 
 var (
+	// Version and BuildCommit can be overridden at build time with:
+	// go build -ldflags="-X github.com/konveyor/crane/internal/buildinfo.Version=<version> -X github.com/konveyor/crane/internal/buildinfo.BuildCommit=$(git rev-parse HEAD)"
 	Version string = "v0.0.6"
 
 	CranelibVersion string = cranelibversion.Version
+	BuildCommit     string = ""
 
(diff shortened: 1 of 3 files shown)


internal/buildinfo/buildinfo.go
```diff
@@ -10,9 +10,12 @@ import (
 )
 
 var (
+	// Version and BuildCommit can be overridden at build time with:
+	// go build -ldflags="-X github.com/konveyor/crane/internal/buildinfo.Version=<version> -X github.com/konveyor/crane/internal/buildinfo.BuildCommit=$(git rev-parse HEAD)"
 	Version string = "v0.0.6"
```
aufi: Small nit, could we default to `latest` or `main`?

aufi (approved):
Looks great, thanks for the PR! Left one minor comment.

internal/buildinfo/buildinfo.go
msajidmansoori12: @aufi Thanks, good point. I’ll set the default BuildCommit to main (used only when ldflags are not provided), while CI/tag builds will continue injecting the real commit SHA.

msajidmansoori12:
@coderabbit summary

PR #493, opened 2026-06-09
Fix fail fast

### LINKED ISSUE #471: [BUG] transform should fail fast when export directory is missing (without creating stage directories)
## Description

When running:

`./crane transform -p .local-plugins`

and the `export/` directory does not exist, `crane transform` currently:

1. Discovers no existing stages
2. Creates default stage directories (for example `transform/10_KubernetesPlugin` and `transform/15_OpenShiftPlugin`)
3. Then fails with an error about missing export input

This leaves partial/stale artifacts in `transform/` even though the command failed.  
In fail-fast workflows, this is confusing and can affect follow-up runs by making stage discovery stateful after a failed invocation.

## Expected behavior

- `transform` should validate input preconditions first (including export directory existence/readability).
- If the export directory is missing, it should fail immediately with a clear error.
- No stage directories or other transform artifacts should be created on this failure path.

Seeing similar behaviour for crane apply , if transform dir is not present it still creates output dirs then fails. This should also be fixed.

msajidmansoori12:
## Summary by CodeRabbit

* **Bug Fixes**
  * Validate now rejects missing or non-directory transform-dir and export-dir with clear errors
  * Configuration issues are detected during validation (before execution); validation no longer creates output/transform directories on failure

* **Tests**
  * Added tests covering directory path validation and ensuring failed validation does not create directories

### DIFF
--- a/cmd/apply/apply.go
+++ b/cmd/apply/apply.go
@@ -45,7 +45,16 @@ func (o *Options) Complete(c *cobra.Command, args []string) error {
 }
 
 func (o *Options) Validate() error {
-	// No validation needed - stages are resolved in run()
+	info, err := os.Stat(o.TransformDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("transform-dir %q does not exist", o.TransformDir)
+		}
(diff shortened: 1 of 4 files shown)


aufi (approved):
LGTM

PR #505, opened 2026-06-10
Fix bug 487: plugin-manager add should download correct binary accord…

### LINKED ISSUE #487: [BUG]  `crane plugin-manager add` downloads wrong platform binary,  always picks linux/amd64
### Summary

  `crane plugin-manager add <plugin>` always downloads the first binary in the plugin's version manifest (`Binaries[0]`), which is `linux/amd64` in the upstream index. It ignores `runtime.GOOS` and `runtime.GOARCH`, resulting in an unusable binary on macOS (both Intel and Apple Silicon).

  ### Steps to Reproduce

  ```bash
  # On macOS ARM (Apple Silicon)
  $ go env GOOS GOARCH
  darwin
  arm64

  $ crane plugin-manager add OpenShiftPlugin
  pluginBinary OpenShiftPlugin added to the path - /Users/user/.local/share/crane/plugins

  $ file ~/.local/share/crane/plugins/OpenShiftPlugin
  ELF 64-bit LSB executable, x86-64 ...
  # ^^^ Linux binary on macOS — cannot execute
```

 ### Where the Problem Surfaces

  The downloaded binary sits in ~/.local/share/crane/plugins/. When the user runs crane transform, crane scans this directory and tries to execute each plugin binary. A linux binary on macOS:

  1. crane transform silently fails or errors — the plugin binary can't execute (exec format error), so the OpenShift-specific transformations never run
  2. User gets no clear error — depending on how crane handles plugin execution failures, it may silently skip the plugin or produce a cryptic error
  3. User thinks the pipeline works — transform completes with only the KubernetesPlugin, missing all OpenShift-specific whiteouts (builder/deployer SAs, system RoleBindings, openshift-service-ca.crt, pull secrets)
  4. Apply to target produces warnings and duplicates — all the OCP system resources that should have been stripped end up in the output

### PR DESCRIPTION
**Pull request by stillalearner**
# PR: Fix bug 487: plugin-manager add downloads wrong platform binary

  Fixes [#487](https://github.com/migtools/crane/issues/487). `crane plugin-manager add` always downloaded `linux/amd64` binary regardless of the user's platform.

  ## Root Cause

  `FilterPluginForOsArch` used a value-based range loop (`for _, version := range`), so the filtered binary was never persisted back to the struct. Additionally, `add.go` hardcoded `Binaries[0]` instead of selecting by platform.

  ## The Fix

  - Fixed `FilterPluginForOsArch` to use index-based iteration so the filter persists
  - Replaced `Binaries[0]` with `binaryURIForPlatform()` that matches `runtime.GOOS`/`runtime.GOARCH` and errors clearly when no binary exists for the current platform

  ## Manual Testing

  | Test | Before | After |
  |------|--------|-------|
  | `crane plugin-manager add OpenShiftPlugin` on darwin/arm64 | `ELF 64-bit x86-64` (wrong) | `Mach-O 64-bit arm64` (correct) |
  | `crane plugin-manager add ImageStreamPlugin` on darwin/arm64 | `ELF 64-bit x86-64` (wrong) | `Mach-O 64-bit arm64` (correct) |
  | Downloaded plugin executes | `exec format error` | Runs correctly |

  ## Files Changed

  | File | Change |
  |------|--------|
  | `internal/plugin/plugin_manager_helper.go` | Fixed range loop to persist filter |
  | `cmd/plugin-manager/add/add.go` | Platform-aware binary selection with clear error |
  | `internal/plugin/plugin_manager_helper_test.go` | 2 tests for filter persistence and no-match case |

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin installation now correctly selects the appropriate version for your operating system and CPU architecture rather than defaulting to a generic option.
  * Enhanced error handling and messaging when a plugin version is unavailable for your platform.

### DIFF
--- a/cmd/plugin-manager/add/add.go
+++ b/cmd/plugin-manager/add/add.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"runtime"
 	"syscall"
 
 	"github.com/konveyor/crane/internal/flags"
@@ -155,7 +156,11 @@ func (o *Options) run(args []string) error {
 				for _, value := range pluginsMap[args[0]] {
 					// check if the version is mentioned and matches the version in pluginsMap file
 					if value.Name != "" && (o.Version == "" || string(value.Version) == o.Version) {
-						return downloadBinary(o.PluginDir, value.Name, value.Binaries[0].URI, log)
+						uri, err := binaryURIForPlatform(value)
+						if err != nil {
+							return err
+						}
+						return downloadBinary(o.PluginDir, value.Name, uri, log)
 					} else {
 						log.Errorf("The version %s of plugin %s is not available", installVersion, value.Name)
 						fmt.Printf("Run \"crane plugin-manager list --name %s --params\" to see available versions along with additional information \n", args[0])
@@ -173,7 +178,11 @@ func (o *Options) run(args []string) error {
 				}
 				for _, value := range pluginsMap[args[0]] {
 					if string(value.Version) == installVersion {
-						return downloadBinary(o.PluginDir, value.Name, value.Binaries[0].URI, log)
+						uri, err := binaryURIForPlatform(value)
+						if err != nil {
+							return err
+						}
+						return downloadBinary(o.PluginDir, value.Name, uri, log)
 					}
 				}
 				log.Errorf("The %s version of the plugin %s is not found", installVersion, args[0])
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by aufi — APPROVED** [review]
LGTM, confirmed available binaries and looks OK 👍

File: cmd/plugin-manager/add/add.go [hunk 1]
```diff
@@ -238,3 +247,14 @@ func downloadBinary(filepath string, filename string, url string, log *logrus.Lo
 	log.Infof("pluginBinary %s added to the path - %s", filename, filepath)
 	return err
 }
+
+// binaryURIForPlatform returns the download URI for the current OS/arch.
+// Returns an error if no matching binary is available.
+func binaryURIForPlatform(version plugin.PluginVersion) (string, error) {
```
midays, line 253: `binaryURIForPlatform` doesn't have a unit test. Since it's a pure function and it's the direct fix for the Binaries[0] issue, it seems like a natural place to pin the platform-selection logic, I think it's worth adding a test for it here, WDYT ?

midays (changes requested):
LGTM, left one comment

cmd/plugin-manager/add/add.go
stillalearner: Sure, I will add a unit test for this function, in following PR, as this one i merged already.

Thanks. a lot ! good point!

PR #514, opened 2026-06-11
fix label selector issue for cluster scoped resources

### LINKED ISSUE #470: [BUG] Label selector filters out cluster-scoped RBAC resources, breaking SA permission chain
### Summary

  When using `crane export -l <selector>`, the label selector is applied to ALL resource types including cluster-scoped resources (ClusterRoleBindings, ClusterRoles). Since users typically don't label RBAC resources with
  app-level labels, the cluster-scoped List calls return empty results and the RBAC relationship matching logic (`filterRbacResources`) never runs. ServiceAccounts are exported but their associated RBAC permissions are silently
  dropped.

  CRDs are **not affected** — they are fetched by name based on which custom resource types were found in the export, not listed with the label selector.

  ### Affected Resources

  | Resource Type | How it's discovered | Label selector applied? | Affected? |
  |--------------|---------------------|------------------------|-----------|
  | ClusterRoleBinding | Listed directly via API | **Yes** | **Yes — breaks RBAC chain** |
  | ClusterRole | Listed directly via API | **Yes** | **Yes — breaks RBAC chain** |
  | SCC (OpenShift) | Listed directly via API | **Yes** | **Yes — same issue** |
  | CRD | `Get` by name, derived from exported CRs | No | No |

  ### Impact

  - ServiceAccounts are exported (they have the app label), but their ClusterRoleBindings and ClusterRoles are silently dropped
  - After `transform` + `apply` on the target cluster, the SA exists with **no RBAC permissions**
  - Application fails on the target with permission errors
  - No warning or indication that RBAC resources were skipped

  ### Steps to Reproduce

  ```bash
  # Setup: SA has app=mysql label, CRB and CR have no labels
  kubectl create sa crane-test-sa -n crane-test
  kubectl label sa crane-test-sa app=mysql -n crane-test

  kubectl create clusterrole crane-test-reader --verb=get,list --resource=pods
  kubectl create clusterrolebinding crane-test-reader-binding \
    --clusterrole=crane-test-reader \
    --serviceaccount=crane-test:crane-test-sa

  # Export with label selector
  crane export -n crane-test -l app=mysql

  # Result: SA exported, ClusterRoleBinding and ClusterRole missing
  ls export/resources/crane-test/
  # ServiceAccount__v1_crane-test_crane-test-sa.yaml  ← present
  # No ClusterRole or ClusterRoleBinding files         ← missing

  # Without label selector, RBAC is correctly exported
  crane export -n crane-test
  ls export/resources/crane-test/_cluster/
  # ClusterRole_..._crane-test-reader.yaml             ← present
  # ClusterRoleBinding_..._crane-test-reader-binding.yaml ← present

stillalearner:
## Fix

  - Fix label selector being applied to cluster-scoped RBAC resources (ClusterRoleBindings, ClusterRoles, SCCs) during `crane export -l <selector>`, which silently dropped the entire SA permission chain
  - Add unit test verifying label selector is not passed to the API for cluster-scoped resources

  ## Problem

  When using `crane export -l <label>`, the label selector was applied to **all** resource types including cluster-scoped resources. Since users typically don't label RBAC resources with app-level labels (e.g.
  `app=mysql`), the cluster-scoped List calls returned empty results and `filterRbacResources` never ran. ServiceAccounts were exported but their associated ClusterRoleBindings and ClusterRoles were silently dropped,
  causing permission errors on the target cluster.

  Test plan

  - [x] Unit test TestGetObjects_clusterScopedSkipsLabelSelector confirms label selector is not passed for cluster-scoped resources
  - [x] Unit test TestGetObjects_passesLabelSelectorToList confirms label selector is still applied for namespaced resources
  - [x] go build ./... passes
  - [x] go test ./... passes (all unit tests)
  - [x] Manual verification on minikube (src context): crane export -n crane-test -l app=mysql correctly exports SA + ClusterRoleBinding + ClusterRole

  Fixes: #470 

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected label selector handling for cluster-scoped Kubernetes resources to prevent incorrect filtering behavior.

* **Tests**
  * Added test coverage to verify proper label selector handling across different resource types.

### DIFF
--- a/cmd/export/discover.go
+++ b/cmd/export/discover.go
@@ -332,7 +332,7 @@ func getObjects(g *groupResource, namespace string, labelSelector string, d dyna
 		}
 	})
 	listOptions := metav1.ListOptions{}
-	if labelSelector != "" {
+	if labelSelector != "" && g.APIResource.Namespaced {
 		listOptions.LabelSelector = labelSelector
 	}
 
--- a/cmd/export/discover_test.go
+++ b/cmd/export/discover_test.go
@@ -1230,6 +1230,36 @@ func TestGetObjects_passesLabelSelectorToList(t *testing.T) {
 	}
 }
 
+func TestGetObjects_clusterScopedSkipsLabelSelector(t *testing.T) {
+	crb := &unstructured.Unstructured{}
+	crb.SetAPIVersion("rbac.authorization.k8s.io/v1")
+	crb.SetKind("ClusterRoleBinding")
+	crb.SetName("test-binding")
+	client := dynamicfake.NewSimpleDynamicClient(runtime.NewScheme(), crb)
+	var sawLabel string
+	client.PrependReactor("list", "clusterrolebindings", func(action kubetesting.Action) (handled bool, ret runtime.Object, err error) {
+		la, ok := action.(kubetesting.ListActionImpl)
+		if !ok {
+			t.Fatalf("expected ListActionImpl, got %T", action)
+		}
+		sawLabel = la.GetListOptions().LabelSelector
+		return false, nil, nil
+	})
+	g := &groupResource{
+		APIGroup:        "rbac.authorization.k8s.io",
+		APIVersion:      "v1",
+		APIGroupVersion: "rbac.authorization.k8s.io/v1",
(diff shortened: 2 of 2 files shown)


aufi (approved):
LGTM

PR #515, opened 2026-06-11
fix: make `crane export` fail non-zero when all namespace list calls are Forbidden

### LINKED ISSUE #480: [BUG] `export` should fail (non-zero) when all namespace resource listings are Forbidden
## Description

When running `crane export` with a namespace-scoped non-admin user against an inaccessible/nonexistent namespace, the command currently:

- logs many `cannot list obj in namespace ...` errors
- writes failure artifacts under `<export-dir>/failures/`
- exports no actual resources (`<export-dir>/resources/` is empty)
- still exits with code `0`

This creates a false-success signal for users and CI pipelines, because the command appears successful even though export effectively failed.

### Example context

- Command:
  - `./crane export -n does-not-exist-xyz --context src-sajid -e exportMissingA`
- Observed:
  - repeated Forbidden/cannot-list messages
  - `resources/` empty, `failures/` populated
  - exit code `0`

## Expected behavior

If export cannot list any resource types for the target namespace (effectively all return Forbidden / inaccessible), the command should:

1. Exit with non-zero status (`exit 1`)
2. Print a clear, actionable summary error

Suggested message:

`export failed: all resource types returned Forbidden for namespace <ns> — verify the namespace exists and your user has list permissions`

## Why this matters

- Prevents false positives in automation
- Makes failure mode obvious and actionable for users
- Aligns exit code semantics with actual export outcome

msajidmansoori12:
## Summary

- Fixes the false-success path in `crane export` where all resource list calls fail with RBAC `Forbidden`, but command still exits `0`.
- Adds strict detection helper in `cmd/export/export.go`:
  - `allResourceListsForbidden(resources, resourceErrs)`
  - returns true only when:
    - no resources were exported, and
    - at least one list error exists, and
    - every list error is `Forbidden`.
- Updates export run flow to:
  - continue writing failure artifacts under `<export-dir>/failures/<namespace>/`
  - then return a command error so CLI exits non-zero.
- Adds regression unit tests:
  - `TestAllResourceListsForbidden` in `cmd/export/export_test.go`
  - covers all-forbidden, mixed errors, no-errors, and resources-exported scenarios.

## Impact

- **Severity:** Medium (behavioral fix for exit code semantics in a specific failure mode).
- **Affected components:** `cmd/export/export.go`, `cmd/export/export_test.go`.
- **User/CI impact:** Prevents false positives in automation when export is fully blocked by namespace RBAC.
- **Compatibility:** Successful and partial-success export paths remain unchanged.

## Test plan

- [x] `go test ./cmd/export -v`
- [x] `go run main.go export -n export-test --context src-sajid`
- [x] Verified final error message is actionable and command exits non-zero (`exit status 1`)
- [x] Verified strict condition message:
  - `export failed: all resource types returned Forbidden for namespace "<ns>" -- verify the namespace exists and your user has list permissions`

### `go run main.go export -n export-test --context src-sajid`

```text
WARN[0000] cannot verify namespace "export-test" exists (may lack RBAC permission): namespaces "export-test" is forbidden: User "sajid" cannot get resource "namespaces" in API group "" in the namespace "export-test"
ERRO[0000] cannot list obj in namespace for groupVersion v1, kind: Service
ERRO[0000] cannot list obj in namespace for groupVersion v1, kind: ServiceAccount
ERRO[0000] cannot list obj in namespace for groupVersion v1, kind: Pod
ERRO[0000] cannot list obj in namespace for groupVersion apps/v1, kind: Deployment
ERRO[0000] cannot list obj in namespace for groupVersion rbac.authorization.k8s.io/v1, kind: ClusterRoleBinding
ERRO[0000] cannot list obj in namespace for groupVersion operators.coreos.com/v1alpha1, kind: Subscription
Error: export failed: all resource types returned Forbidden for namespace "export-test" -- verify the namespace exists and your user has list permissions
...
exit status 1
```

## Notes

- This PR intentionally keeps per-resource failure details in output for diagnosis and still surfaces a single final actionable command-level failure.
- Namespace existence pre-check warning is preserved; final export failure is determined from actual list outcomes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved export error reporting when no resources are discovered because all resource list attempts were forbidden; messages now include the target namespace and guidance to verify namespace existence and list permissions.

* **Tests**
  * Added test coverage for permission-related export failure scenarios to ensure correct detection and messaging.

### DIFF
--- a/cmd/export/export.go
+++ b/cmd/export/export.go
@@ -202,7 +202,6 @@ func (o *ExportOptions) Run() error {
 	var errs []error
 
 	resources, resourceErrs := resourceToExtract(o.userSpecifiedNamespace, o.labelSelector, dynamicClient, resourceLists, log)
-	allForbidden := allResourceListsForbidden(resources, resourceErrs)
 	clusterScopeHandler := NewClusterScopeHandler()
 	resources = clusterScopeHandler.filterRbacResources(resources, log)
 
@@ -237,9 +236,9 @@ func (o *ExportOptions) Run() error {
 
 	errs = append(errs, writeResourcesErrors...)
 	errs = append(errs, writeErrorsErrors...)
-	if allForbidden {
+	if allResourceListsForbidden(resources, resourceErrs) {
 		errs = append(errs, fmt.Errorf(
-			"export failed: all resource types returned Forbidden for namespace %q -- verify the namespace exists and your user has list permissions",
+			"all resource types returned Forbidden for namespace %q -- verify the namespace exists and your user has list permissions",
 			o.userSpecifiedNamespace,
 		))
 	}
--- a/cmd/export/export.go
+++ b/cmd/export/export.go
@@ -117,6 +117,25 @@ func validateExportNamespace(ctx context.Context, client kubernetes.Interface, n
 	return nil
 }
 
+// allResourceListsForbidden reports whether export discovered no resources and
+// every per-resource list attempt failed with a Forbidden error. This is used
+// to trigger a non-zero exit for the strict "all Forbidden" export failure case.
+func allResourceListsForbidden(resources []*groupResource, resourceErrs []*groupResourceError) bool {
+	if len(resources) > 0 || len(resourceErrs) == 0 {
+		return false
+	}
+
+	for _, resourceErr := range resourceErrs {
+		if resourceErr == nil || resourceErr.Error == nil {
+			return false
+		}
+		if !apierrors.IsForbidden(resourceErr.Error) {
+			return false
+		}
+	}
+	return true
+}
+
 // Run performs discovery, lists resources, filters cluster-scoped RBAC to related
 // ServiceAccounts, writes YAML under exportDir, and returns an aggregate of non-fatal write errors.
 func (o *ExportOptions) Run() error {
@@ -183,6 +202,7 @@ func (o *ExportOptions) Run() error {
 	var errs []error
 
 	resources, resourceErrs := resourceToExtract(o.userSpecifiedNamespace, o.labelSelector, dynamicClient, resourceLists, log)
+	allForbidden := allResourceListsForbidden(resources, resourceErrs)
 	clusterScopeHandler := NewClusterScopeHandler()
 	resources = clusterScopeHandler.filterRbacResources(resources, log)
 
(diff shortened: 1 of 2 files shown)


cmd/export/export.go
```diff
@@ -217,7 +237,12 @@ func (o *ExportOptions) Run() error {
 
 	errs = append(errs, writeResourcesErrors...)
 	errs = append(errs, writeErrorsErrors...)
-
+	if allForbidden {
+		errs = append(errs, fmt.Errorf(
+			"export failed: all resource types returned Forbidden for namespace %q -- verify the namespace exists and your user has list permissions",
```
istein1: **claude:**
Error message wording — Minor suggestion

  The error message reads:

  ▎ export failed: all resource types returned Forbidden for namespace "X" -- 
  ▎ verify the namespace exists and your user has list permissions

  Since Run() is called from RunE, cobra will prefix this with Error:, so users
  will see:

  ▎ Error: export failed: all resource types returned Forbidden...

  Consider dropping the export failed: prefix to avoid redundancy — the error is
  already surfaced by the export command. But this is cosmetic.

cmd/export/export.go
```diff
@@ -183,6 +202,7 @@ func (o *ExportOptions) Run() error {
 	var errs []error
 
 	resources, resourceErrs := resourceToExtract(o.userSpecifiedNamespace, o.labelSelector, dynamicClient, resourceLists, log)
+	allForbidden := allResourceListsForbidden(resources, resourceErrs)
```
istein1: **claude**

Minor issue: allForbidden is computed too early relative to resourceErrs 
  mutations

  At line 185 (new code), allForbidden is computed. But then at line 192,
  crdErrs are appended to resourceErrs:
  resources, resourceErrs := resourceToExtract(...)
  allForbidden := allResourceListsForbidden(resources, resourceErrs)  // line 
  185
  ...
  crdErrs := ...
  resourceErrs = append(resourceErrs, crdErrs...)  // line 192
  
  This means CRD collection errors are excluded from the all-Forbidden check. In
  practice this is likely fine — if all namespace list calls are Forbidden,
  there are no resources to trigger CRD lookups, so crdErrs will be empty. But
  it's worth a comment explaining the intentional ordering, or moving the check
  after the CRD append for robustness.

aufi: I personally would suggest not use the variable here and just call the function on line 240 (this could look little misleading that there is something else that handles or works with `allForbidden`). But just a nit, not blocking.

aufi (approved):

cmd/export/export.go
msajidmansoori12: Thanks for the review Ilanit , fixed it.

msajidmansoori12: @aufi thanks for the review , agreed , I have fixed it.

midays (approved):
LGTM

PR #525, opened 2026-06-12
Fix dependency error in resources ordering

### LINKED ISSUE #266: [BUG] kubectl apply on crane output directory fails on first run due to alphabetical file ordering
### Description:
When applying crane's output directory to a fresh target namespace, kubectl apply -f <output-dir>/ processes files alphabetically. This causes RoleBinding resources that reference a Role to be applied before the Role itself exists on the target, resulting in a NotFound error on the first apply.

### Steps to Reproduce:

**1. Ensure a clean environment:**

```
kubectl delete namespace role-migration --context src --ignore-not-found=true
kubectl delete namespace role-migration --context tgt --ignore-not-found=true
```

**2. Set up a fresh target namespace with non-admin access:**

```
kubectl create namespace role-migration --context tgt
kubectl create rolebinding nonadmin-ns-admin \
  --clusterrole=admin \
  --user=dev \
  -n role-migration \
  --context tgt
```

**3. Set up source namespace with a Role and RoleBinding:**
```
kubectl create namespace role-migration --context src
kubectl create serviceaccount test-sa -n role-migration --context src
kubectl create role pod-reader --verb=get,list,watch --resource=pods -n role-migration --context src
kubectl create rolebinding pod-reader-binding --role=pod-reader --serviceaccount=role-migration:test-sa -n role-migration --context src
```

**4. Run the crane pipeline:**

```
EXPORT_DIR=$(mktemp -d)
TRANSFORM_DIR=$(mktemp -d)
OUTPUT_DIR=$(mktemp -d)

crane export --namespace role-migration --context src --export-dir "$EXPORT_DIR"
crane transform --export-dir "$EXPORT_DIR" --transform-dir "$TRANSFORM_DIR"
crane apply --export-dir "$EXPORT_DIR" --transform-dir "$TRANSFORM_DIR" --output-dir "$OUTPUT_DIR"
```

**5. Apply the output to the fresh target namespace in a single pass:**

```
KUBECONFIG=~/.kube/config \
kubectl apply \
  -f "$OUTPUT_DIR/resources/role-migration/" \
  --context tgt-dev
```

### Actual Result:
```
rolebinding.rbac.authorization.k8s.io/nonadmin-ns-admin configured
role.rbac.authorization.k8s.io/pod-reader created
serviceaccount/test-sa created
Error from server (NotFound): error when creating "...RoleBinding_rbac.authorization.k8s.io_v1_role-migration_pod-reader-binding.yaml": rolebindings.rbac.authorization.k8s.io "pod-reader" not found
```

### Expected Result:
```
All resources apply successfully in a single pass.
Root Cause:
kubectl apply -f <dir> processes files alphabetically. RoleBinding sorts before Role, so pod-reader-binding is attempted before pod-reader exists on the target. Kubernetes validates the roleRef at creation time and rejects the binding.
```

### Workaround:
Run kubectl apply again, the second pass succeeds because all dependencies are already in place:

`kubectl apply -f "$OUTPUT_DIR/resources/role-migration/" --context tgt`

### Suggested Fix:
crane should write manifests to the output directory in dependency-aware order, ensuring that Role resources are written before RoleBinding resources, ConfigMap/Secret before Deployment, etc.

### Environment:
 
crane version: v0.0.6
crane-lib version: v0.0.10
Kubernetes: minikube v1.35.1
kubectl: v1.33.3

aufi:
Fixing alphabetical file ordering issue where RoleBinding was applied before Role, causing NotFound errors on first kubectl apply.

**OPT-IN with `--ordered` flag for `crane apply` **

Changes:
- Add 3-digit order prefix to resource filenames (e.g., 300_Role_*, 310_RoleBinding_*)
- Implement dependency-aware ordering following kubectl best practices
- Resources now apply in correct order: Namespace → Secret/ConfigMap → Role → RoleBinding → Deployment → Service

Files are now named with order prefixes ensuring dependencies exist before dependents:
```
  240_ConfigMap_*
  300_Role_*
  310_RoleBinding_*
  340_Deployment_*
```

This eliminates the need to run kubectl apply twice and follows Kubernetes resource ordering conventions.

Fixes #266

## Summary by CodeRabbit

* **New Features**
  * Added an `--ordered` option to the apply command to write Kubernetes resources with dependency-aware filename prefixes for more reliable apply sequencing.
* **Tests**
  * Added/expanded unit tests validating resource ordering logic, filename prefix formatting, lexicographic stability, and correct ordering for workloads and webhook resources.


stillalearner:
So, this PR fixes the resource ordering issue while applying all the files inside the output/resources/ns/  directory. 
While there is another file  by kustomise in the output/ directory, named as output.yaml, which also contains all the resources, and ordered correctly as well, which can be directly applied by the user. 

So what we need to decide first is the correct way of applying all the resources at once on the target cluster:

1. kubectl apply -f output/output.yaml 

-> This is already sorted appropriately by kustomize according to the dependency of resources. Then we wont be needing this fix.

2. kubectl apply -f output/resources/ns/ 

-> This is not sorted acc to dependency, AND this fix would be needed in this case.

aufi:
Updated to OPT-IN for numbering and added `--ordered` flag for apply.

internal/file/ordering.go
```diff
@@ -0,0 +1,94 @@
...
+	if order, exists := ResourceOrder[kind]; exists {
+		return order
+	}
+	return ResourceOrder["_default"]
+}
+
+// GetOrderedResourceFilename returns a filename with an order prefix for dependency-aware application.
+// Format: NNN_Kind_group_version_namespace_name.yaml
+// where NNN is a 3-digit order number (e.g., "050_Role_rbac.authorization.k8s.io_v1_default_pod-reader.yaml")
+func GetOrderedResourceFilename(obj unstructured.Unstructured) string {
```
msajidmansoori12, line 92: Hi @aufi , 
just one small thing I noticed that might be worth considering.

In --ordered mode, filenames are built with 3-digit padding (%03d), but unknown kinds use order 1000 (4 digits).
So you can end up with names like:

1000_UnknownKind_...
240_ConfigMap_...
300_Role_...
When sorted as text, 1000_... comes before 240_..., which is opposite of the intended “unknown kinds go last” ordering.

This may not always impact normal kubectl apply -f <dir> usage, so totally up to you whether to address now or later. I mainly wanted to flag it since it can affect deterministic filename ordering (especially for scripts/tools that apply files in sorted-name order).

Possible future tweak (if you decide to): switch to 4-digit padding (e.g. %04d), so ordering is consistent: 0240_..., 0300_..., 1000_....

Currently : 
```
 ls | sort 
1000_UnknownKind_example.com_v1_default_u1.yaml
240_ConfigMap__v1_default_cm1.yaml
```

If using 4 digit padding - 

```
ls | sort                                                                
0240_ConfigMap__v1_default_cm1.yaml
1000_UnknownKind_example.com_v1_default_u1.yaml
240_ConfigMap__v1_default_cm1.yaml
```

msajidmansoori12 (changes requested):

internal/file/ordering.go
aufi: You're right, thanks for reporting this.

aufi: I updated the default 1000 to 999 and kept 3 digits to stay concistent (and I believe 0-999 will be enough at least for now).

msajidmansoori12: Thanks @aufi

stillalearner (approved):
LGTM, Thanks for the opt-in feature. @aufi

midays (approved):
LGTM

internal/file/ordering.go
aufi: Thanks @msajidmansoori12 , could you please re-review? (To remove change request)

msajidmansoori12 (approved):

midays:
the PR fixes the issue:

generated files are ordered with `--ordered` flag
```
ls ./output-test/resources/role-migration
220_ServiceAccount__v1_role-migration_test-sa.yaml
300_Role_rbac.authorization.k8s.io_v1_role-migration_pod-reader.yaml
310_RoleBinding_rbac.authorization.k8s.io_v1_role-migration_pod-reader-binding.yaml
```

no error or warning is printed:
```
kubectl apply \
  -f "$OUTPUT_DIR/resources/role-migration/" \
  --context tgt-dev

serviceaccount/test-sa created
role.rbac.authorization.k8s.io/pod-reader created
rolebinding.rbac.authorization.k8s.io/pod-reader-binding created
```

Issue #479, opened 2026-06-08
[BUG] crane transform shows confusing <nil> in error message for invalid stage name


RanWurmbrand:
**What version of crane are you running, and what are your clutsters+platform**
crane built from source (main branch), Fedora 44

**What did you expect to happen?**
A clear error message explaining why the stage name is invalid.

**What actually happened?**
The error message includes <nil> instead of an explanation: invalid custom stage name "my stage": <nil>

**Please include any relevant logs or errors**

/migration-wordpress$ crane transform "my stage"
INFO[0000] Stage for "my stage" not found, attempting to create 
Error: invalid custom stage name "my stage":  `<nil>`

PR #527, opened 2026-06-14
fix: resolve nil error in custom stage name validation

Tamar-Dinavetsky:
[#479](https://github.com/migtools/crane/issues/479)

### Description
Fixes a bug where providing an invalid custom stage name to the `crane transform` command returned a `<nil>` error instead of a helpful error message.

### What was wrong?
In the original implementation, the variable `err` was scoped inside the `if` block checking for successful validation (`if err := validateStageNameToken(requested); err == nil`). When the validation actually failed, the code skipped the `if` block and returned `fmt.Errorf("...", err)`. However, at that point, the outer function's `err` variable was shadowed and evaluated to `<nil>`.

### What was fixed?
Refactored the validation check using a clean guard clause (`if validationErr != nil`) to properly capture and bubble up the actual inner validation error.

### Testing & Verification
Verified locally by running:
`crane transform "my stage" -e ./test-export -t ./test-transform`

- Before: `Error: invalid custom stage name "my stage": <nil>`
- After: `Error: invalid custom stage name "my stage": contains unsafe characters (only A-Z, a-z, 0-9, -, _ allowed)`
## Summary by CodeRabbit

* **Refactor**
  * Internal improvements to error handling and validation logic for custom stage creation. No user-visible changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for custom stage names so invalid names fail cleanly and prevent unintended stage directory creation.
* **Tests**
  * Added coverage for invalid custom stage name inputs and strengthened assertions around returned error behavior.
* **Chores**
  * Updated ignore rules to exclude additional local artifacts and temporary project directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,10 +17,3 @@ vendor/
 
 crane
 .DS_Store
-
-# My Personal Tools and Scripts
-scripts/ship_it.sh
-scripts/mac_refresh.py
-slack-creds-extractor-main/
-crane-test/
-.claude/
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,10 @@ vendor/
 
 crane
 .DS_Store
+
+# My Personal Tools and Scripts
+scripts/ship_it.sh
+scripts/mac_refresh.py
+slack-creds-extractor-main/
+crane-test/
+.claude/
--- a/cmd/transform/transform.go
+++ b/cmd/transform/transform.go
@@ -733,18 +733,18 @@ func (o *Options) resolveAndValidateStages(
 				continue
 			}
 
-			if err := validateStageNameToken(requested); err == nil {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -16,4 +16,11 @@
 vendor/
 
 crane
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# My Personal Tools and Scripts
```
aufi: The project `.gitignore` should not contain user-specific local files, please move it from this PR to your local setup (examples e.g. at https://dev.to/fronkan/a-personal-gitignore-even-for-a-single-repository-4o7h)

aufi (approved):
Thanks for the PR! The fix itself looks great, but please check comment on `.gitignore` before merging.

.gitignore
Tamar-Dinavetsky: Removed personal files from .gitignore. Ready for merge, thank you!

Issue #415, opened 2026-05-27
Plugin system foundational update for new resources creation


aufi:
This is a higher level issue to track this effort. It makes sense to update at least crane-lib part soon to be ready for test day to fully utilize multistage transformation with non-trivial plugins.

Goal: Enable Crane plugins to generate new Kubernetes resources (not just transform existing ones). 

### Current State
The Crane plugin system allows plugins to only:

Transform existing resources using JSONPatch operations
Mark resources for deletion using the whiteout flag
Limitation: Cannot create new resource types that didn't exist in the input data.

### Requirement
Enable plugins to generate new resources for scenarios like:

BuildConfig → Shipwright Build - OpenShift to Kubernetes-native conversion
DeploymentConfig → Deployment - deprecated resource migration
ImageStream → container registry references - OpenShift dependency resolution

### Architecture
```
Input Resources
    ↓
[Orchestrator] → [Runner] → [Plugin]
    ↓                           ↓
RunnerResponse.NewResources ←─┘
    ↓
TransformArtifacts (original + new)
    ↓
[Writer] → Kustomize structure
```

Key Insight: The existing Writer already handles artifacts uniformly, no changes needed!

Details on https://github.com/aufi/move-crane/blob/main/drafts/plugin-update-new-resource-plan.md

PR #545, opened 2026-06-17
Deprecate plugin optionals

### LINKED ISSUE #437: Consolidate plugins optionals arg
In `crane transform`, there is an inherited option to pass optional args to plugins https://github.com/migtools/crane/blob/main/cmd/transform/optionals/optionals.go#L50, example:

```
# manually installed OpenShift and ImageStream plugins as addition to KubernetesPlugin to see maximum of possible optional flags
$ crane transform optionals
Plugin: KubernetesPlugin (version v0.0.10)
    add-annotations: Annotations to add to each resource
        Example: annotation1=value1,annotation2=value2
    registry-replacement: Map of image registry paths to swap on transform, in the format original-registry1=target-registry1,original-registry2=target-registry2...
        Example: docker-registry.default.svc:5000=image-registry.openshift-image-registry.svc:5000,docker.io/foo=quay.io/bar
    remove-annotations: Annotations to remove
        Example: annotation1,annotation2
    disable-whiteout-owned: Disable whiting out owned pods and pod template resources
        Example: true
    extra-whiteouts: Additional resources to whiteout specified as a comma-separated list of GroupKind strings.
        Example: Deployment.apps,Service,Route.route.openshift.io
    include-only: If specified, every resource not listed here will be a whiteout. extra-whiteouts is ignored when include-only is specified. Specified as a comma-separated list of GroupKind strings.
        Example: Deployment.apps,Service,Route.route.openshift.io
    strip-default-rbac: Whether to strip default RBAC including default serviceAccount (default: true)
        Example: true
    strip-default-cabundle: Whether to strip default CA Bundle (default: true)
        Example: true
    pvc-rename-map: A comma-separated list of colon separated pvc renames.
        Example: old-pvc1-name:new-pvc1-name,old-pvc2-name:new-pvc2-name
Plugin: ImagestreamPlugin (version v0.0.2)
    src-internal-registry: Internal registry hostname[:port] used to determine whether an istag references a local image
        Example: image-registry.openshift-image-registry.svc:5000
Plugin: OpenShiftPlugin (version v0.0.4)
    strip-default-rbac: Whether to strip default RBAC including builder and deployers serviceAccounts, roleBindings for admin, builders, and deployers (default: true)
        Example: true
    strip-default-cabundle: Whether to strip default CA Bundle (default: true)
        Example: true
    strip-default-pull-secrets: Whether to strip Pod and BuildConfig default pull secrets (beginning with builder/default/deployer-dockercfg-) that aren't replaced by the map param pull-secret-replacement (default: true)
        Example: true
    pull-secret-replacement: Map of pull secrets to replace in Pods and BuildConfigs while transforming in format secret1=destsecret1,secret2=destsecret2[...]
        Example: default-dockercfg-h4n7g=default-dockercfg-12345,builder-dockercfg-abcde=builder-dockercfg-12345
    registry-replacement: Map of image registry paths to swap on transform, in the format original-registry1=target-registry1,original-registry2=target-registry2...
        Example: docker-registry.default.svc:5000=image-registry.openshift-image-registry.svc:5000,docker.io/foo=quay.io/bar
    pvc-rename-map: A comma-separated list of colon separated pvc renames.
        Example: old-pvc1-name:new-pvc1-name,old-pvc2-name:new-pvc2-name
```

This was not yet re-considered how this should look like in multistage transformation. Probably the main question I think needs to answered is if the KubernetesPlugin with default options together with possibility have custom stages (based on kustomize) can  provide the same or more features than optional flags for legacy plugins. Or if there could be good use-case for other plugins (e.g. like passing fragment of kustomization together with a custom stage creation).

In the end, this could be a bug either
- there is an `optionals` CLI arg, that is redundant/misleading/.. and should be removed
- there is an `optionals` CLI arg, that needs to be fixed to correctly work with multistage transformation (update multistage transform (and instructions file) to allow passing these options for each stage (it can be different for each stage))

aufi:
The original crane plugin system supports optional flags to be passed to plugins. Continuation of these flags was described in https://github.com/migtools/crane/issues/437 with aim to remove it.

However, to stay on safe side, pending crane plugin system update (partially related to #415) and some gut feeling about possible need for args for each transform stage, I chose the way to mark these flags as deprecated, but keeping them for v0.10 release.

Change
1. Marked related flags and subcommands as deprecated in cobra
2. Added small test to check they work (so, we'll see a failure after removing them)

Fixes: https://github.com/migtools/crane/issues/437

### DIFF
--- a/cmd/transform/optionals/optionals.go
+++ b/cmd/transform/optionals/optionals.go
@@ -47,8 +47,9 @@ func NewOptionalsCommand(f *flags.GlobalFlags) *cobra.Command {
 		cobraGlobalFlags: f,
 	}
 	cmd := &cobra.Command{
-		Use:   "optionals",
-		Short: "Return a list of optional fields accepted by configured plugins",
+		Use:        "optionals",
+		Short:      "Return a list of optional fields accepted by configured plugins",
+		Deprecated: "use custom stages with kustomization patches instead. Optional flags apply globally to all stages and will be removed in a future version.",
 		RunE: func(c *cobra.Command, args []string) error {
 			if err := o.Complete(c, args); err != nil {
 				return err
--- a/cmd/transform/optionals_test.go
+++ b/cmd/transform/optionals_test.go
@@ -1,7 +1,6 @@
 package transform
 
 import (
-	"encoding/json"
 	"testing"
 
 	jsonpatch "github.com/evanphx/json-patch"
@@ -10,17 +9,14 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-// mockPluginWithOptionals implements cranelib.Plugin and tracks whether optionals were received
+// mockPluginWithOptionals implements cranelib.Plugin and captures received optionals
 type mockPluginWithOptionals struct {
 	name              string
 	receivedOptionals map[string]string
 }
 
 func (m *mockPluginWithOptionals) Run(request cranelib.PluginRequest) (cranelib.PluginResponse, error) {
-	// Capture the optionals that were passed to the plugin
 	m.receivedOptionals = request.Extras
-
-	// Return a simple response with no transformations (empty patch)
 	return cranelib.PluginResponse{
 		Version:    string(cranelib.V1),
 		Patches:    jsonpatch.Patch{},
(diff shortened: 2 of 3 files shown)


stillalearner (approved):
lgtm

PR #548, opened 2026-06-18
improved log for custom stage

### LINKED ISSUE #477: [BUG] [TRANSFORM] Silently creates empty directories and reports success for non-existent stage names
## Summary
When running `crane transform` with a completely non-existent or misspelled stage name (e.g., `99_FakeStage`), Crane does not validate the input against registered plugins or fail-fast. Instead, it silently creates an empty custom stage directory, executes with 0 plugins applied, copies the resources as-is, and returns a successful exit status (Exit Code 0). 

This behavior is highly misleading for users, as typographical errors can go unnoticed, giving a false impression that a specific transformation logic was executed successfully.

## Steps to Reproduce
1. Generate an export directory with standard Kubernetes manifests.
2. Run the transform command providing a completely fabricated stage name:
```bash
   ./crane transform 99_FakeStage -e ./export-dir -t ./transform-dir
```
3. Observe the terminal output and the generated directory structure.

## Expected Behavior
The CLI should validate the provided stage name. If the stage does not match any registered plugins or predefined configurations, it should either:

- Fail-fast and return an error (`Exit Code 1`).

- Or, at minimum, print a prominent warning message (e.g., WARN: `No plugins matched "99_FakeStage". Creating an empty custom stage directory.`) to ensure the user is aware that no active code was executed.

##Actual Behavior / Output Log
The command completes with a success status and prints the following log:

```
INFO[0000] Stage for "99_FakeStage" not found, attempting to create 
INFO[0000] Creating custom stage directory: 99_FakeStage 
INFO[0000] Running 1 stage(s): [99_FakeStage]           
INFO[0000] Executing stage 1/1: 99_FakeStage            
INFO[0000] Stage 99_FakeStage: loaded 8 input resource(s) 
INFO[0000] Stage 99_FakeStage: produced 8 output resource(s) 
INFO[0000] Successfully completed 1 stage(s)    
```
        
An empty 99_FakeStage directory is generated under the transform path, and resources are copied over without any active mutation.

## Environment
OS: MacOS
Crane Version: v0.10.0-alpha.1

stillalearner:
Improved log for custom stage. 

Fixes #477 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved custom stage creation messages when no plugins are involved, clarifying that resources will be copied as-is and advising to delete the directory and re-run if the stage name may have been mistyped.

### DIFF
--- a/cmd/transform/transform.go
+++ b/cmd/transform/transform.go
@@ -569,7 +569,7 @@ func (o *Options) createCustomStageWithExplicitName(
 	transformDir string,
 	log *logrus.Logger,
 ) error {
-	log.Warnf("Creating custom stage directory %q with no plugins — resources will be copied as-is. If this was a typo, delete the directory and re-run with the correct stage name.", stageName)
+	log.Infof("Creating custom stage directory %q with no plugins — resources will be copied as-is. If this was a typo, delete the directory and re-run with the correct stage name.", stageName)
 
 	if err := o.ensurePreviousStagesRun(orchestrator, transformDir, log); err != nil {
 		return fmt.Errorf("failed to ensure previous stages are run: %w", err)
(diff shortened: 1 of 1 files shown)


cmd/transform/transform.go
```diff
@@ -572,7 +569,7 @@ func (o *Options) createCustomStageWithExplicitName(
 	transformDir string,
 	log *logrus.Logger,
 ) error {
-	log.Infof("Creating custom stage directory: %s", stageName)
+	log.Warnf("Creating custom stage directory %q with no plugins — resources will be copied as-is. If this was a typo, delete the directory and re-run with the correct stage name.", stageName)
```
aufi: There should stay `Info` IMO (custom stage is a standard option to be used), let's discuss it during a call later today.

stillalearner: Ack, DONE

RanWurmbrand:
LGTM

RanWurmbrand (approved):

Issue #382, opened 2026-05-14
[BUG] Make validate output more readable when suggestions are made


nachandr:
**What version of crane are you running, and what are your clutsters+platform**

**What did you expect to happen?**

**What actually happened?**
1) The output currently has 2 columns (among others) : `Reason`, `Suggestion`.
    The string in the `Reason` column also includes the suggestion in brackets. This should be removed.
2) Additionally, if the strings in the `REASON` column could be shortened, that would be great. The output looks jumbled up on small screens.
```
$ crane validate -i ./test-suggestions --context crane-dest-context
INFO[0000] Scanned 6 distinct GVK+namespace tuples      
Mode: live (context: crane-dest-context)

APIVERSION           KIND                     NAMESPACE           RESOURCE  STATUS        REASON                                                                                                         SUGGESTION                                  
autoscaling/v2beta2  HorizontalPodAutoscaler  multi-resource-app            Incompatible  API version autoscaling/v2beta2 not available on target cluster (available as autoscaling/v1, autoscaling/v2)  available as autoscaling/v1, autoscaling/v2  
batch/v1beta1        CronJob                  multi-resource-app            Incompatible  API version batch/v1beta1 not available on target cluster (available as batch/v1)                              available as batch/v1                        
extensions/v1beta1   Deployment               multi-resource-app            Incompatible  API version extensions/v1beta1 not available on target cluster (available as apps/v1)                          available as apps/v1                         
extensions/v1beta1   Ingress                  multi-resource-app            Incompatible  API version extensions/v1beta1 not available on target cluster (available as networking.k8s.io/v1)             available as networking.k8s.io/v1            
extensions/v1beta1   NetworkPolicy            multi-resource-app            Incompatible  API version extensions/v1beta1 not available on target cluster (available as networking.k8s.io/v1)             available as networking.k8s.io/v1            
policy/v1beta1       PodDisruptionBudget      multi-resource-app            Incompatible  API version policy/v1beta1 not available on target cluster (available as policy/v1)                            available as policy/v1                       

Summary: 6 scanned, 0 compatible, 6 incompatible
Result: FAILED — 6 resource(s) incompatible with target cluster
INFO[0000] Wrote validation report to validate/report.json 
INFO[0000] Wrote 6 validation failure(s) to validate/failures 
Error: validation failed: one or more resources are incompatible with the target cluster
```

**Please include any relevant logs or errors**

nachandr:
[...]

PR #565, opened 2026-06-22
fix: resolve duplicate suggestion text in validate output reason column

Tamar-Dinavetsky:
[#382](https://github.com/migtools/crane/issues/382)

## 📝 Summary

This PR fixes an issue where the output of `crane validate` duplicated the suggestion text inside the `REASON` column, causing jumbled and overly wide tables on smaller screens. 

### 🕵️‍♂️ Root Cause
In `internal/validate/matcher.go`, the `addSuggestion` function was responsible for both setting `result.Suggestion` and appending the same text to `result.Reason`. This second operation was completely redundant since the suggestion already had its own dedicated field and column in the output table.

### 🛠️ Fix Applied
- Removed the line that appended the suggestion to `result.Reason` in `internal/validate/matcher.go`.
- The `REASON` column now correctly contains only the reason for incompatibility, while the `SUGGESTION` column shows the alternatives.

### 🧪 Test Changes
- Updated `TestMatchResults_SuggestionWhenAlternativeExists` in `internal/validate/matcher_test.go` to remove the assertion that checked if `Reason` contains "available as", as this was validating the old incorrect behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,13 @@ vendor/
 
 crane
 .DS_Store
+
+
+# My Personal Tools and Scripts
+scripts/ship_it.sh
+scripts/mac_refresh.py
+scripts/slack.py
+slack-creds-extractor-main/
+crane-test/
+.claude/
+.env
--- a/.gitignore
+++ b/.gitignore
@@ -17,13 +17,3 @@ vendor/
 
 crane
 .DS_Store
-
-
-# My Personal Tools and Scripts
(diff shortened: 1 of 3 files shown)


RanWurmbrand:
LGTM

RanWurmbrand (approved):

nachandr (approved):

stillalearner (commented):
LGTM

Issue #481, opened 2026-06-08
[BUG] KUBECONFIG with multiple files breaks `crane export` and `crane validate`


stillalearner:
### Summary

  Kubernetes supports multiple kubeconfig files via the `KUBECONFIG` environment variable (colon-separated paths). This works with `kubectl` and with crane's `transform`/`apply` commands, but fails in `crane export` and `crane validate`.

  ### What happens

  ```bash
  # This is standard Kubernetes usage — works everywhere
  export KUBECONFIG=~/.kube/config-cluster1:~/.kube/config-cluster2

  # kubectl works fine
  kubectl --context cluster1 get pods   ✅

  # crane export fails
  crane export --context cluster1 -n my-app
  # Error: stat /home/user/.kube/config-cluster1:/home/user/.kube/config-cluster2: no such file or directory

 ```

 Crane treats the entire colon-separated string as a single file path instead of splitting it into two files and merging them.

 ### Steps to Reproduce

  1. Set KUBECONFIG with multiple paths (standard Kubernetes practice)
  2. export KUBECONFIG=$HOME/.kube/config:$HOME/.kube/config

  ### Run crane export
  ./crane export --context src -n crane-test

  ### Expected: export succeeds (kubectl works fine with the same KUBECONFIG)
  ### Actual: Error: stat /path/a:/path/b: no such file or directory

  -> Workaround

  Merge your kubeconfig files into a single file, or avoid the KUBECONFIG env var and use the --kubeconfig flag pointing to a single file

PR #573, opened 2026-06-24
fix: support multi-file KUBECONFIG in crane export and validate

Tamar-Dinavetsky:
[#481](https://github.com/migtools/crane/issues/481)
## Problem

  Kubernetes supports specifying multiple kubeconfig files via the
  KUBECONFIG environment variable, separated by colons:

 ```
 export KUBECONFIG=~/.kube/config-cluster1:~/.kube/config-cluster2
 ```
This works with kubectl and with crane's transform/apply commands,
but crane export and crane validate failed with:
 ```
Error: stat /home/user/.kube/config-cluster1:/home/user/.kube/config-cluster2: no such file or directory
```

  ## Root Cause

  crane uses viper with `AutomaticEnv() `enabled globally. When
  `viper.BindPFlags() `registers the `--kubeconfig` flag, viper automatically
  reads the KUBECONFIG environment variable and maps its raw value
  (the full colon-separated string) into configFlags.KubeConfig.

  When `configFlags.ToRawKubeConfigLoader()` is later called, it sets
  `loadingRules.ExplicitPath` to that raw colon-separated string.
  `os.Stat()`is then called on the full string as a single path,
  which does not exist.

  transform and apply are not affected because their Complete()
  functions do not connect to the cluster at all.

  ## Fix

  At the start of Complete() in both export and validate, reset
  configFlags.KubeConfig to an empty string if the user did not
  explicitly pass --kubeconfig on the command line.

  This prevents viper from injecting the raw KUBECONFIG env var
  into the loading rules, allowing NewDefaultClientConfigLoadingRules()
  to correctly split KUBECONFIG by colon and load each file separately —
  the same behavior as kubectl.

  ## Files Changed

  - cmd/export/export.go
  - cmd/validate/validate.go

  ## Testing

  Verified manually with two kubeconfig files:

      export KUBECONFIG=~/.kube/config:~/.kube/config-copy
      crane export --context minikube -n default   ✅
      crane validate -i export/resources/default --context minikube   ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how command-line operations handle kubeconfig when run programmatically, making behavior more predictable.
  * Validation and export commands now consistently treat an unspecified kubeconfig as empty instead of inheriting implicit defaults.

### DIFF
--- a/cmd/export/export.go
+++ b/cmd/export/export.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd/api"
+
 )
 
 // ExportOptions holds CLI flags and runtime state for a single export run.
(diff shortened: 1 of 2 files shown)


stillalearner (approved):
LGTM

nachandr (approved):

PR #574, opened 2026-06-24
feat(e2e): add --run-as flag to run all tests as cluster-admin

### LINKED ISSUE #467: add option to run test as both admin and non-admin

midays:
Add a --run-as=admin flag that overrides all test contexts to use
cluster-admin credentials without modifying individual test files.

Changes:
- config: add RunAs field
- suite: register --run-as flag in tier0 and tier1 suite files
- scenario: populate NonAdmin fields with admin contexts when RunAs=admin
- rbac: add SetupActiveKubectlRunners and SetupActiveNamespaceAdmin
  helpers that bypass RBAC setup in admin mode and log the resolved
  username as proof of identity
- tests: remove empty-context Skip guards from all 21 non-admin tests
  and replace SetupNamespaceAdminUsersForScenario with
  SetupActiveKubectlRunners

When --run-as is omitted, behavior is unchanged.

**[Jenkins Run](https://jenkins-csb-migrationqe-main.dno.corp.redhat.com/job/mta/job/crane-ocp-tests-runner/213/testReport/)**

## Summary by CodeRabbit

* **New Features**
  * Added a `--run-as` flag to run end-to-end tests with admin credentials.
* **Bug Fixes**
  * Removed multiple premature skips so non-admin E2E migrations and validations run even when non-admin contexts are not provided.
  * Updated non-admin kubectl/permission initialization to consistently follow the correct admin vs non-admin execution path across tiers and scenarios.
* **Tests**
  * Added a compliance workflow to enforce non-admin test conventions around setup and skip guards.

### DIFF
--- a/e2e-tests/framework/rbac.go
+++ b/e2e-tests/framework/rbac.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"strings"
 
+	"github.com/konveyor/crane/e2e-tests/config"
 	"github.com/konveyor/crane/e2e-tests/utils"
 )
 
@@ -106,6 +107,46 @@ func SetupNamespaceAdminUser(adminKubectl KubectlRunner, nonAdminContext, namesp
 	}
 
 	return userKubectl, cleanup, nil
+}
+
+// SetupActiveNamespaceAdmin sets up namespace-scoped access on a single cluster.
+// When config.RunAs == "admin", it creates the namespace with the admin runner and returns
+// a no-op cleanup. Otherwise it delegates to SetupNamespaceAdminUser.
+func SetupActiveNamespaceAdmin(adminKubectl KubectlRunner, nonAdminContext, namespace string) (KubectlRunner, func(), error) {
+	if config.RunAs == "admin" {
+		if err := adminKubectl.CreateNamespace(namespace); err != nil {
+			return KubectlRunner{}, nil, fmt.Errorf("failed to create namespace %q: %w", namespace, err)
+		}
+		return adminKubectl, func() {}, nil
+	}
+	return SetupNamespaceAdminUser(adminKubectl, nonAdminContext, namespace)
+}
+
+// SetupActiveKubectlRunners returns kubectl runners appropriate for the current RunAs mode.
+// When config.RunAs == "admin", it creates the namespace with admin credentials and returns
+// admin runners directly (bypassing RBAC setup). Otherwise it delegates to
+// SetupNamespaceAdminUsersForScenario to grant namespace-scoped permissions to the non-admin user.
+func SetupActiveKubectlRunners(scenario MigrationScenario, namespace string) (KubectlRunner, KubectlRunner, func(), error) {
+	if config.RunAs == "admin" {
+		if srcUser, err := ResolveUsernameForContext(scenario.KubectlSrc.Context); err != nil {
+			log.Printf("[run-as=admin] source context=%q: could not resolve username: %v", scenario.KubectlSrc.Context, err)
+		} else {
+			log.Printf("[run-as=admin] source context=%q running as user=%q", scenario.KubectlSrc.Context, srcUser)
+		}
+		if tgtUser, err := ResolveUsernameForContext(scenario.KubectlTgt.Context); err != nil {
+			log.Printf("[run-as=admin] target context=%q: could not resolve username: %v", scenario.KubectlTgt.Context, err)
+		} else {
+			log.Printf("[run-as=admin] target context=%q running as user=%q", scenario.KubectlTgt.Context, tgtUser)
+		}
+		if err := scenario.KubectlSrc.CreateNamespace(namespace); err != nil {
+			return KubectlRunner{}, KubectlRunner{}, nil, fmt.Errorf("failed to create namespace %q on source: %w", namespace, err)
+		}
+		if err := scenario.KubectlTgt.CreateNamespace(namespace); err != nil {
+			return KubectlRunner{}, KubectlRunner{}, nil, fmt.Errorf("failed to create namespace %q on target: %w", namespace, err)
+		}
+		return scenario.KubectlSrc, scenario.KubectlTgt, func() {}, nil
+	}
+	return SetupNamespaceAdminUsersForScenario(scenario, namespace)
 }
 
 // SetupNamespaceAdminUsersForScenario grants namespace-scoped admin permissions
--- a/.github/workflows/test-compliance.yml
+++ b/.github/workflows/test-compliance.yml
@@ -1,0 +1,58 @@
+name: Test Compliance
+
+on:
+  pull_request:
+    paths:
+      - 'e2e-tests/tests/**'
+
+permissions:
+  contents: read
+
+jobs:
+  run-as-admin-compliance:
+    name: run-as-admin compliance
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check for non-compliant test patterns
+        run: |
+          set +e
+          FAILED=0
+
+          # Check all test files in the repo
+          CHANGED=$(find e2e-tests/tests -name '*_test.go')
+
+          echo "Checking files:"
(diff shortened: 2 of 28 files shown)


stillalearner (approved):
LGTM

aufi (approved):

istein1:
**Claude Review**: PR #574 — feat(e2e): add --run-as flag to run all tests as cluster-admin

  This PR adds a --run-as=admin flag that overrides non-admin e2e test contexts to use cluster-admin credentials. It introduces two new helpers
  (SetupActiveKubectlRunners, SetupActiveNamespaceAdmin) that bypass RBAC setup in admin mode, removes empty-context Skip guards from 21 test files, and adds a CI
  compliance workflow to enforce the new patterns. The admin-mode path is clean and well-structured. However, the removal of Skip guards introduces a behavioral
  regression in the non-admin, non-run-as path.

  **Findings**

  1. e2e-tests/tests/tier0/mta_804_empty_pvc_migration_test.go:23 (and 18 other test files) — Removing Skip guards changes behavior from graceful Skip to test Failure 
  when non-admin contexts aren't provided (affects all 19 tests that call SetupActiveKubectlRunners)

  1. When --run-as is omitted AND --source-nonadmin-context/--target-nonadmin-context are also omitted, NewMigrationScenario sets KubectlSrcNonAdmin.Context = "".
  SetupActiveKubectlRunners delegates to SetupNamespaceAdminUsersForScenario, which calls SetupNamespaceAdminUser at rbac.go:64 where nonAdminContext == "" triggers
  fmt.Errorf("non-admin context is required"). Previously, the Skip guards caught this and tests showed as Skipped (green). Now they Fail (red). The PR description
  claims "When --run-as is omitted, behavior is unchanged" — this is incorrect for environments that don't provide non-admin contexts. Consider adding an early Skip
  inside SetupActiveKubectlRunners when RunAs == "" and the non-admin context is empty, preserving the original Skip semantics in one place.
  
** The fix is small: add an empty-context Skip inside SetupActiveKubectlRunners so
  the guard lives in one place instead of 19 test files.
  
  2. .github/workflows/test-compliance.yml:28 — Compliance workflow misses non-admin tests that don't call SetupActiveKubectlRunners

  2. The grep filter SetupActiveKubectlRunners\|SetupNamespaceAdminUsersForScenario skips any file not containing these symbols. Tests mta_846 and mta_850 had their Skip
  guards removed in this PR but use scenario.CraneNonAdmin / scenario.KubectlTgtNonAdmin.Context directly — they escape the compliance check entirely. Future non-admin
  tests following this pattern will also go undetected. Consider matching on NonAdmin field usage (e.g. CraneNonAdmin\|NonAdmin) as an additional detection signal.
  3. e2e-tests/tests/tier0/mta_846_validate_single_incompatible_route_live_test.go:32 — Skip guard removed but test uses empty non-admin context directly without any 
  setup or fallback

  3. Unlike the other 19 tests, mta_846 doesn't call SetupActiveKubectlRunners. After Skip guard removal, when --target-nonadmin-context is empty,
  scenario.KubectlTgtNonAdmin.Context is "" which gets passed to runner.Validate(ValidateOptions{Context: ""}). The Validate function omits --context when empty, so
  crane silently falls back to the default kubeconfig context (likely admin). The test won't crash, but it silently runs as admin instead of non-admin — the opposite of
  what the test intends. The same issue applies to mta_850.
  
  **Findings importance**
  Finding 1 is the one that matters. It's a real behavioral regression: tests that used to Skip now Fail in any environment that doesn't provide non-admin contexts,
  contradicting the PR's own claim that behavior is unchanged without --run-as.

  Findings 2 and 3 are minor — they're about two validate-only tests where the non-admin distinction barely matters (they test API compatibility, not RBAC). 
  Finding 4  (no flag validation) is trivial.

e2e-tests/framework/rbac.go
```diff
@@ -108,6 +109,46 @@ func SetupNamespaceAdminUser(adminKubectl KubectlRunner, nonAdminContext, namesp
...
+		return adminKubectl, func() {}, nil
+	}
+	return SetupNamespaceAdminUser(adminKubectl, nonAdminContext, namespace)
+}
+
+// SetupActiveKubectlRunners returns kubectl runners appropriate for the current RunAs mode.
+// When config.RunAs == "admin", it creates the namespace with admin credentials and returns
+// admin runners directly (bypassing RBAC setup). Otherwise it delegates to
+// SetupNamespaceAdminUsersForScenario to grant namespace-scoped permissions to the non-admin user.
+func SetupActiveKubectlRunners(scenario MigrationScenario, namespace string) (KubectlRunner, KubectlRunner, func(), error) {
```
msajidmansoori12, line 129: Hi @midays  — I found one behavior change while testing this PR locally.

I checked out the PR branch and ran a focused tier0 test (MTA-810) without passing non-admin contexts (and without --run-as=admin). It now fails at setup with:

non-admin context is required

So in default mode, this path is now fail instead of skip. Previously these non-admin tests would skip when non-admin contexts were missing.

Could we restore that previous skip behavior when --run-as != admin and non-admin contexts are not provided? That would keep “when --run-as is omitted, behavior is unchanged” true.

Please let me know if there was a strong reason for failing these tests deliberately then I'll retract this comment. 

Thanks.

```
------------------------------
ConfigMap Migration [MTA-810] Should migrate a configmap with data intact as nonadmin user [tier0]
/Users/mmansoor/Desktop/git_repos/crane-testing/crane/.worktrees/pr574-review/e2e-tests/tests/tier0/mta_810_configmap_test.go:16
  STEP: Grant ns admin permissions to nonadmin user on source and target @ 07/01/26 16:29:48.801
  [FAILED] in [It] - /Users/mmansoor/Desktop/git_repos/crane-testing/crane/.worktrees/pr574-review/e2e-tests/tests/tier0/mta_810_configmap_test.go:41 @ 07/01/26 16:29:48.801

[CRANE-RESULT] FAILED MTA-810
• [FAILED] [0.001 seconds]
ConfigMap Migration [It] [MTA-810] Should migrate a configmap with data intact as nonadmin user [tier0]
/Users/mmansoor/Desktop/git_repos/crane-testing/crane/.worktrees/pr574-review/e2e-tests/tests/tier0/mta_810_configmap_test.go:16

  [FAILED] Unexpected error:
      <*errors.errorString | 0x14000326470>: 
      non-admin context is required
      {
          s: "non-admin context is required",
      }
  occurred
  In [It] at: /Users/mmansoor/Desktop/git_repos/crane-testing/crane/.worktrees/pr574-review/e2e-tests/tests/tier0/mta_810_configmap_test.go:41 @ 07/01/26 16:29:48.801
```

msajidmansoori12 (changes requested):

e2e-tests/framework/rbac.go
midays: Hi, thanks for the detailed feedback and for testing this locally.

The FAIL behavior is intentional. These tests were built specifically to run with non-admin context, so if the context isn't provided, that's a misconfiguration of the test run rather than an expected condition, so failing loudly is more helpful than a silent skip in that case.

Also, the Jenkins pipeline always passes `--source-nonadmin-context` and `--target-nonadmin-context` regardless of the `RUN_AS_ADMIN` setting, so there's no scenario in a real pipeline run where these contexts would be missing.

WDYT ?

msajidmansoori12: Right , makes sense @midays , thanks for clearing it up.

msajidmansoori12 (approved):

midays:
@istein1 thank you for sharing it, i've updated PR to fix accordingly, please note regarding the first one:
```
e2e-tests/tests/tier0/mta_804_empty_pvc_migration_test.go:23 (and 18 other test files) — Removing Skip guards changes behavior from graceful Skip to test Failure
when non-admin contexts aren't provided (affects all 19 tests that call SetupActiveKubectlRunners)
```
this is intentional and has been addressed: https://github.com/migtools/crane/pull/574#discussion_r3505788679

istein1 (approved):

PR #586, opened 2026-06-26
drop empty maps in comparison

stillalearner:
Dropping empty maps for comparison of golden manifests. 

It causes issue while comparing the outputs on minikube vs ocp

### DIFF
--- a/e2e-tests/utils/utils.go
+++ b/e2e-tests/utils/utils.go
@@ -814,7 +814,13 @@ func normalizeWithPath(value any, path []string) any {
 				continue
 			}
 			nextPath := append(append([]string{}, path...), k)
-			out[k] = normalizeWithPath(child, nextPath)
+			normalized := normalizeWithPath(child, nextPath)
+			// Drop empty maps — an empty map {} is semantically identical to
+			// the field not existing and should not cause comparison failures.
+			if m, ok := normalized.(map[string]any); ok && len(m) == 0 {
+				continue
+			}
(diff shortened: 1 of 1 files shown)


e2e-tests/utils/utils.go
```diff
@@ -814,7 +821,13 @@ func normalizeWithPath(value any, path []string) any {
...
 			}
 			nextPath := append(append([]string{}, path...), k)
-			out[k] = normalizeWithPath(child, nextPath)
+			normalized := normalizeWithPath(child, nextPath)
+			// Drop empty maps — an empty map {} is semantically identical to
+			// the field not existing and should not cause comparison failures.
+			if m, ok := normalized.(map[string]any); ok && len(m) == 0 {
+				continue
+			}
+			out[k] = normalized
```
msajidmansoori12, line 832: @stillalearner We already use targeted normalization in this file (shouldDropField / normalizeUnstableFields), so can we follow the same pattern and only treat empty securityContext as optional at the known paths (pod spec / pod template spec) , as from what I understand for the time being that was what is causing the test to fail ? 

A global drop can probably hide real differences in other fields and make tests pass when they shouldn’t.

msajidmansoori12 (changes requested):

msajidmansoori12 (approved):

e2e-tests/utils/utils.go
stillalearner: ack

PR #592, opened 2026-07-01
:test_tube: crane validate offline mode : Automate test for malformed API surface file handling

### LINKED ISSUE #380: Test and automate: (tier1) Offline mode : Verify behavior with malformed API surface JSON file

nachandr:
Fixes [#380 ](https://github.com/migtools/crane/issues/380)

* **Tests**
  * Added tier1 end-to-end coverage for `[MTA-860]` offline validation when the API surface input file contains malformed or mis-shaped JSON..
  * Confirms no validation report is produced for malformed inputs.

### DIFF
--- a/e2e-tests/tests/tier1/mta_860_validate_malformed_api_surface_test.go
+++ b/e2e-tests/tests/tier1/mta_860_validate_malformed_api_surface_test.go
@@ -1,0 +1,363 @@
+package e2e
+
+import (
+	"encoding/json"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/konveyor/crane/e2e-tests/config"
+	. "github.com/konveyor/crane/e2e-tests/framework"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Crane validate offline mode: malformed API surface file handling", func() {
+	It("[MTA-860] Should handle malformed API surface JSON file gracefully as namespace admin",
+		Label("tier1", "validate", "offline"), func() {
+		appName := "multi-resource-app"
+		namespace := "validate-malformed-json"
+
+		scenario := NewMigrationScenario(
+			appName,
+			namespace,
+			config.K8sDeployBin,
+			config.CraneBin,
+			config.SourceContext,
+			config.TargetContext,
+		)
+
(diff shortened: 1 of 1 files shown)


e2e-tests/tests/tier1/mta_860_validate_malformed_api_surface_test.go
```diff
@@ -0,0 +1,252 @@
...
+					matchers := make([]types.GomegaMatcher, len(tc.errorSubstrings))
+					for idx, substr := range tc.errorSubstrings {
+						matchers[idx] = ContainSubstring(substr)
+					}
+					Expect(rootErrMsg).To(Or(matchers...), "root error message should indicate expected issue")
+
+					// Verify validation report was not created
+					reportPath := filepath.Join(validateDir, "report.json")
+					Expect(reportPath).NotTo(BeAnExistingFile(), "report.json should not be created with malformed JSON")
+				} else {
```
msajidmansoori12: Hi @nachandr , 

One small question on test intent: in the current cases, every entry sets expectError: true, so the success branch (else) is never exercised. Would you like this spec to stay strictly negative-path coverage? If yes, we could simplify by removing expectError/else to make intent explicit. If you want mixed coverage, maybe add one valid API-surface case so the success path is also tested.

Either direction works — mostly suggesting this for clarity and future maintenance.

msajidmansoori12 (changes requested):

e2e-tests/tests/tier1/mta_860_validate_malformed_api_surface_test.go
nachandr: > Hi @nachandr ,
> 
> One small question on test intent: in the current cases, every entry sets expectError: true, so the success branch (else) is never exercised. Would you like this spec to stay strictly negative-path coverage? If yes, we could simplify by removing expectError/else to make intent explicit. If you want mixed coverage, maybe add one valid API-surface case so the success path is also tested.
> 
> Either direction works — mostly suggesting this for clarity and future maintenance.

Hi @msajidmansoori12 , 
This test is strictly for malformed scenarios since the valid API surface scenarios are being covered by other offline tests. I have removed the success branch else block.

msajidmansoori12 (approved):
lgtm

PR #594, opened 2026-07-01
fix nginx test

stillalearner:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how resource identities are matched for export comparisons, making generated names more stable.
  * ReplicaSet entries now use the same owner-based identity handling as Pods, reducing false differences when names include generated suffixes.


midays (approved):
LGTM

PR #599, opened 2026-07-02
:test_tube: Add input validation for --run-as flag

nachandr:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for the E2E `run-as` setting so invalid values stop the test run early.
  * Improved startup logging to clearly show whether tests are running in admin or non-admin mode.

The --run-as flag accepts any string, but only "admin" has special behavior. Typos like --run-as=admon or --run-as=adming  do nothing  and fall back to non-admin mode and the test is run as non-admin.  Users get unexpected behavior without any error message

  Example of the problem:
  $ User makes a typo
 ```$ ginkgo run -v --focus="\[MTA-850\]" e2e-tests/tests/tier1 --   --k8sdeploy-bin=~/k8s-apps-deployer/venv/bin/k8sdeploy   --crane-bin=~/migtools-crane/crane/crane   --source-context=kind-source-cluster   --target-context=kind-dest-cluster   --source-nonadmin-context=crane-source-context   --target-nonadmin-context=crane-dest-context --run-as=admon```

  Solution:
  Fail fast with a clear error message:
  Invalid --run-as value: "admon". Valid values: "admin" or empty string (for non-admin mode)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added startup validation for the test run mode so invalid `run-as` values are caught early.
  * Improved test run logging to clearly show whether suites are running in admin or non-admin mode.

### DIFF
--- a/e2e-tests/config/config.go
+++ b/e2e-tests/config/config.go
@@ -1,5 +1,7 @@
 package config
 
+import "log"
+
 var (
 	K8sDeployBin          string
 	SourceContext         string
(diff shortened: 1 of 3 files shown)


midays (approved):
LGTM

PR #603, opened 2026-07-05
:bug: [transfer-pvc] Fix silent data loss on vanilla K8s by reading w…

### LINKED ISSUE #213: [BUG] [transfer-pvc] Permission denied on uid-restricted directories causes silent data loss
## Description

When migrating a stateful application (MongoDB) using crane transfer-pvc, directories on the source PVC owned by an application-specific uid (e.g. uid 999, the mongodb user) with 700 permissions are silently skipped by rsync. The destination pod starts with an empty database, with no clear indication to the user that data was not transferred.

## Steps to reproduce

1. Deploy MongoDB on source cluster via k8s-apps-deployer (or any deployment that runs MongoDB as uid 999)
2. Insert test documents into the database
3. Run crane transfer-pvc without any pre-steps
4. Deploy MongoDB on dest cluster using the transferred PVC
5. Query the database — collection is empty (count: 0)

## Actual behavior

```
Status: Partially failed
Failed files:
  - /mnt/mongo/data/.mongodb [Permission denied (13)]
  - /mnt/mongo/data/_tmp [Permission denied (13)]
  - /mnt/mongo/data/diagnostic.data [Permission denied (13)]
  - /mnt/mongo/data/journal [Permission denied (13)]
```
## Expected behavior

All files on the source PVC are transferred regardless of file ownership. The destination database contains all source data. This is how MTC/Velero handles the same scenario, velero runs with elevated privileges and reads all files regardless of ownership without any user intervention.

## Root cause

The rsync-server and rsync-client pods spawned by crane transfer-pvc run as uid 1000. MongoDB writes its data directories (journal/, _tmp/, diagnostic.data/, .mongodb/) as uid 999 with 700 (owner-only) permissions. rsync running as uid 1000 cannot read these directories and skips them silently.

## Workaround

```
# Run as root on source PVC before transfer-pvc
chmod -R 755 <source-pvc-mount>
```

stillalearner:
## Summary

  Fixes #213 — `transfer-pvc` silently transfers 0 bytes on vanilla Kubernetes when PVC files have restrictive permissions (0700/0600) owned by an app-specific UID (e.g., MySQL UID 27, MongoDB UID 999).

  ### Root cause

  On OCP, `transfer-pvc` reads the namespace UID annotation (`openshift.io/sa.scc.uid-range`) and runs rsync pods as that UID. On vanilla K8s, this annotation doesn't exist, so rsync falls back to the image default UID
  (65534/nobody). rsync as nobody can't read 0700 files owned by the app UID → silent empty transfer.

  ### Fix

  Three-tier UID detection in `getIDsForNamespace()`:

  1. **OCP namespace annotation** (existing) — reads `openshift.io/sa.scc.uid-range`. No behavior change on OCP.
  2. **Workload spec lookup** (new) — searches Deployments, StatefulSets, DaemonSets, ReplicaSets, Jobs, and CronJobs in the namespace for one that mounts the PVC, then reads `runAsUser`/`fsGroup`/`runAsGroup` from its
  pod securityContext. Handles container-level `runAsUser` overriding pod-level.
  3. **PVC file ownership inspection** (new) — creates a temporary busybox pod to `stat` file ownership on the PVC. Covers apps that rely on Dockerfile `USER` without declaring `runAsUser` in the pod spec (e.g.,
  MongoDB).

  Additional changes:
  - rsync **server** pod now sets `RunAsUser`/`RunAsGroup`/`FSGroup` in PodSecurityContext (previously only FSGroup). Ensures files are written with correct ownership on the target.
  - Source-discovered UID is used as fallback for the target side when the target has no OCP annotation and no workload deployed yet.
  - StatefulSet `volumeClaimTemplate` PVC name pattern matching (`<template>-<sts>-<ordinal>`).

  ### What it covers

  | Scenario | UID source | Before | After |
  |---|---|---|---|
  | OCP → OCP | Namespace annotation | Worked | No change |
  | K8s with explicit `runAsUser` (MySQL) | Workload spec | 0 bytes transferred | **Fixed** |
  | K8s with Dockerfile USER (MongoDB) | PVC file inspection | 0 bytes transferred | **Fixed** |
  | K8s → OCP | Workload spec (source) + annotation (target) | 0 bytes transferred | **Fixed** |
  | OCP → K8s | Annotation (source) + workload spec (target) | 0 bytes transferred | **Fixed** |
  | Root app with 0644 files (nginx, redis) | Falls back to image default | Worked | No change |

  ### Tested on

  **Minikube (vanilla K8s) as non-admin:**
  - MySQL (`runAsUser: 27`, `fsGroup: 27`) — 244 MB transferred, DB data intact (100 rows), MD5 match
  - MongoDB (no securityContext, Dockerfile USER 999) — 314 MB transferred, 2 documents intact
  - Container-level `runAsUser` override (pod=1000, container=2000) — 0700 dirs readable, MD5 match
  - StatefulSet with volumeClaimTemplate — PVC name pattern matched, 0700 dirs readable
  - Job with PVC — Job lookup path, 0700 dirs readable
  - Multi-container pod (writer=4001, sidecar=4002) — picks first container UID
  - Redis/nginx (root apps, 0644 files) — no regression
  
    **OCP as non-admin:**
  - MySQL app
  - MongoDB app

  ### Known limitations

  - Multi-UID PVCs with restrictive permissions (e.g., volume written by two different UIDs with 0700) — rsync runs as one UID, can't read the other's files. Requires **admin/root.**
  - File inspection adds ~10-15s (temporary pod lifecycle). Only runs when steps 1 and 2 both fail.
  - Apps running as root with 0700 files — `RunAsNonRoot: true` prevents rsync from running as root. Files must be world-readable for transfer.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PVC transfer reliability by applying more precise security-context settings for the rsync client/server, including safer fallbacks when values aren’t available.
  - Enhanced UID/GID/FSGroup discovery by using namespace configuration, detecting workloads that mount the PVC, and deriving values from PVC content when needed.

- **Tests**
  - Expanded unit tests for PVC reference detection and security-context extraction/precedence.
  - Updated e2e MongoDB and MTA-807 data validation flows by removing the prior PVC permission/ownership workaround steps and helpers.

### DIFF
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -734,8 +734,10 @@ func inspectPVCFileOwnership(c client.Client, namespace string, pvcName string) 
 					Image: "busybox",
 					SecurityContext: func() *corev1.SecurityContext {
 						t, f := true, false
+						nobodyUID := int64(65534)
 						return &corev1.SecurityContext{
 							RunAsNonRoot:             &t,
+							RunAsUser:                &nobodyUID,
 							AllowPrivilegeEscalation: &f,
 							Capabilities: &corev1.Capabilities{
 								Drop: []corev1.Capability{"ALL"},
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -560,6 +560,7 @@ func getIDsForNamespace(c client.Client, namespace string, pvcName string) (*cor
 	// Read securityContext from the workload that owns this PVC.
 	wps, err := getSecurityContextFromWorkload(c, namespace, pvcName)
 	if err != nil {
+		log.Printf("workload security context lookup failed for PVC %s/%s: %v", namespace, pvcName, err)
 		return ps, nil
 	}
 	if wps != nil && wps.RunAsUser != nil {
@@ -571,6 +572,7 @@ func getIDsForNamespace(c client.Client, namespace string, pvcName string) (*cor
 	// the UID that wrote the data.
 	uid, err := inspectPVCFileOwnership(c, namespace, pvcName)
 	if err != nil {
+		log.Printf("PVC file ownership inspection failed for %s/%s: %v", namespace, pvcName, err)
 		return ps, nil
 	}
 	if uid != nil {
@@ -596,7 +598,7 @@ func getSecurityContextFromWorkload(c client.Client, namespace string, pvcName s
 	} else {
 		for _, d := range deployments.Items {
 			if podSpecReferencesPVC(d.Spec.Template.Spec, pvcName) {
-				return extractPodSecurityContext(d.Spec.Template.Spec), nil
+				return extractPodSecurityContext(d.Spec.Template.Spec, pvcName), nil
 			}
 		}
 	}
@@ -610,11 +612,15 @@ func getSecurityContextFromWorkload(c client.Client, namespace string, pvcName s
 	} else {
 		for _, s := range statefulSets.Items {
 			if podSpecReferencesPVC(s.Spec.Template.Spec, pvcName) {
-				return extractPodSecurityContext(s.Spec.Template.Spec), nil
+				return extractPodSecurityContext(s.Spec.Template.Spec, pvcName), nil
 			}
 			for _, vct := range s.Spec.VolumeClaimTemplates {
-				if strings.HasPrefix(pvcName, fmt.Sprintf("%s-%s-", vct.Name, s.Name)) {
-					return extractPodSecurityContext(s.Spec.Template.Spec), nil
+				prefix := fmt.Sprintf("%s-%s-", vct.Name, s.Name)
+				if strings.HasPrefix(pvcName, prefix) {
+					suffix := pvcName[len(prefix):]
+					if _, err := strconv.Atoi(suffix); err == nil {
+						return extractPodSecurityContext(s.Spec.Template.Spec, pvcName), nil
+					}
 				}
 			}
 		}
@@ -629,7 +635,7 @@ func getSecurityContextFromWorkload(c client.Client, namespace string, pvcName s
 	} else {
 		for _, d := range daemonSets.Items {
 			if podSpecReferencesPVC(d.Spec.Template.Spec, pvcName) {
-				return extractPodSecurityContext(d.Spec.Template.Spec), nil
+				return extractPodSecurityContext(d.Spec.Template.Spec, pvcName), nil
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -525,41 +539,246 @@ func getIDsForNamespace(client client.Client, namespace string) (*corev1.Securit
...
+	return &uid, nil
+}
+
+func getRsyncClientPodSecurityContext(c client.Client, namespace string, pvcName string) (*corev1.PodSecurityContext, error) {
+	return getIDsForNamespace(c, namespace, pvcName)
+}
+
+func getRsyncServerPodSecurityContext(c client.Client, namespace string, pvcName string) (*corev1.PodSecurityContext, error) {
+	return getIDsForNamespace(c, namespace, pvcName)
 }
```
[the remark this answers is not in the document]
stillalearner: added

stillalearner:
Tested successfully on OCP, 4 failures, not related to this PR changes.

cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -512,54 +527,333 @@ func getNodeNameForPVC(srcClient client.Client, namespace string, pvcName string
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: namespace,
+		},
+		Spec: corev1.PodSpec{
+			RestartPolicy: corev1.RestartPolicyNever,
+			Containers: []corev1.Container{
+				{
+					Name:  "inspect",
+					Image: "busybox:1.37",
```
aufi: Just to not forget, do we need configure or document this image usage (e.g. for non-upstream use cases?)

aufi (approved):
Thanks for the PR, overall looks great 👍  Adding two comments I got from claude analysis, but  not critical, up to your decision how to handle them.

 ### cmd/transfer-pvc/transfer-pvc.go:108 — Malformed OCP annotation now causes fatal error instead of fallback
  - Scenario: Namespace with malformed openshift.io/sa.scc.uid-range annotation (e.g., "invalid" instead of "1000700000/10000"). Old code returned (nil, nil) and fell back to defaults; new code returns error → log.Fatal() terminates program. A previously-working transfer now aborts
  instead of attempting fallback discovery methods.

###  cmd/transfer-pvc/transfer-pvc.go:426 — Missing exit code validation in PVC ownership inspection
  - Scenario: Inspect pod fails (e.g., PVC mount permission denied, busybox image pull failure) with exit code 1 before writing to termination log. wait.Poll completes successfully (accepts PodFailed), but code only checks Terminated.Message emptiness, not ExitCode. Returns (nil, 
  nil) instead of error → rsync launches with wrong security context → permission denied on actual transfer.

cmd/transfer-pvc/transfer-pvc.go
stillalearner: let me replace it with rsync-transfer image only. No extra image.

PR #617, opened 2026-07-08
 :test_tube: Add e2e test for pipeline determinism test

Tamar-Dinavetsky:
#### Description
Adds an E2E test case to verify that running the full Crane pipeline consecutively on the same namespace produces identical outputs at every stage: `export`, `transform`, and `apply`.

#### How it works
* Runs the `export` ➔ `transform` ➔ `apply` pipeline twice on the same quiesced namespace.
* Compares the results across all stages using semantic YAML verification to ensure determinism and prevent dynamic/mutable fields leakage.

#### Verification Logic
* **Export stage:** `utils.CompareDirectoryYAMLSemanticsExport` - semantic comparison with normalization of dynamic fields; on OpenShift uses `CompareDirectoryYAMLSemanticsExportAllowOptionalOCPOutputDefaults`
* **Transform stage:** `utils.CompareDirectoryYAMLSemanticsUnordered` - semantic comparison that treats top-level YAML arrays as unordered sets, so non-deterministic patch operation ordering in `KubernetesPlugin` does not cause false failures
* **Output stage:** `utils.CompareDirectoryYAMLSemantics` - standard semantic comparison; `output.yaml` is fully deterministic between runs

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end determinism test to ensure the crane export → transform → apply pipeline yields identical results across consecutive runs.
  * Extended YAML comparison coverage with new unit and directory-level tests for unordered matching, including order-insensitive handling of top-level arrays.
* **Bug Fixes**
  * Improved YAML directory validation to be order-independent where appropriate, reducing false differences when YAML top-level list ordering changes.

### DIFF
--- a/e2e-tests/utils/utils.go
+++ b/e2e-tests/utils/utils.go
@@ -230,17 +230,18 @@ func CompareDirectoryYAMLSemantics(goldenDir, gotDir string) error {
 }
 
 // sortTopLevelArray returns a sorted copy of arr by canonical JSON representation of
-  // each element, enabling order-independent comparison of top-level YAML sequences
- func sortTopLevelArray(arr []any) []any {
-      sorted := make([]any, len(arr))
-      copy(sorted, arr)
-      sort.Slice(sorted, func(i, j int) bool {
-          a, _ := json.Marshal(sorted[i])
-          b, _ := json.Marshal(sorted[j])
-          return string(a) < string(b)
-      })
-      return sorted
-  }
+// each element, enabling order-independent comparison of top-level YAML sequences
+func sortTopLevelArray(arr []any) []any {
+	sorted := make([]any, len(arr))
+	copy(sorted, arr)
+	sort.Slice(sorted, func(i, j int) bool {
+		a, _ := json.Marshal(sorted[i])
+		b, _ := json.Marshal(sorted[j])
+		return string(a) < string(b)
+	})
+	return sorted
+}
+
 // compareYAMLFileBytesUnordered parses two YAML inputs and compares their decoded
 // document streams treating arrays as unordered sets. relPath is used only for error context.
 func compareYAMLFileBytesUnordered(relPath string, golden, got []byte) error {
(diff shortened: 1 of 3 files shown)


Tamar-Dinavetsky:
Tested on OCP successfully.

e2e-tests/utils/utils.go
stillalearner: Other than this function call. 

Most of the code in this CompareDirectoryYAMLSemantics looks common with CompareDirectoryYAMLSemanticsUnordered.

Can we take out the common code and use in both CompareDirectoryYAMLSemantics and CompareDirectoryYAMLSemanticsUnordered?

stillalearner (approved):
I think the non-admin contexts should be used

stillalearner (changes requested):
I think this should be a non-admin test.

istein1:
1. MEDIUM: No unit tests for new utility functions. sortTopLevelArray, compareYAMLFileBytesUnordered, and CompareDirectoryYAMLSemanticsUnordered are public utilities
  with non-trivial logic (canonical JSON sorting, array-as-set semantics) but have zero unit tests. The codebase has a strong pattern of unit-testing every utility in
  utils_test.go. These should have table-driven tests covering: identical arrays in different order, nested maps, non-array root documents (no-op path), and empty
  arrays.
  2. LOW: sortTopLevelArray silently discards json.Marshal errors. The sort comparator does a, _ := json.Marshal(sorted[i]). If marshaling fails, both values become
  nil/"", making all failed elements compare equal and producing nondeterministic sort order — defeating the function's purpose. Unlikely in practice (data comes from
  parsed YAML), but should propagate the error or use a fallback.
  3. LOW: Gratuitous deletion of useful doc comment. The PR removes // This is intended for stable outputs (for example transform/output artifacts). from
  CompareDirectoryYAMLSemantics. That comment was accurate and helps future callers decide which comparison function to use.
  4. LOW: Misleading By() step. By("Run crane export/transform/apply pipeline") is placed before runner variable setup, not before the actual pipeline execution. Lines
  79 and 82 already have their own By() for each run.
  5. NIT: Trailing blank line before closing brace in CompareDirectoryYAMLSemanticsUnordered.

Tamar-Dinavetsky:
@istein1  @stillalearner fixed , thanks for the review.

stillalearner (approved):
LGTM

PR #625, opened 2026-07-09
:bug: Fix Route hostname collision for long PVC names on OCP

### LINKED ISSUE #287: [BUG] getRouteHostName truncation can cause Route collisions
### getRouteHostName truncation can cause Route collisions

**What version of crane are you running, and what are your clusters+platform**

Latest main branch. Issue exists in `cmd/transfer-pvc/transfer-pvc.go` line 735.

Each PVC should get a unique Route hostname, even when PVC names are long.

**What actually happened?**

`getRouteHostName()` truncates long name-namespace prefixes to 62 chars using `routeNamePrefix[:62]`. Two PVCs with names that share the same first 62 characters will get the same hostname, causing a Route conflict.

Note: `getValidatedResourceName()` on line 485 already solves this problem using MD5 hashing: `fmt.Sprintf("crane-%x", md5.Sum([]byte(name)))`. The same approach could be applied here.

N/A — discovered via code review.

stillalearner:
Fixes #287

  ### Summary
  
  When transferring PVCs with long names on OCP, `getRouteHostName()` truncates the Route hostname prefix to 62 characters. Two PVCs whose names share the same first 62 characters get identical hostnames, causing a Route conflict and transfer failure.
Fixed by truncating to 53 characters and appending an 8-character MD5 hash suffix, keeping the hostname human-readable while guaranteeing uniqueness within the DNS label limit of 63 characters.

  ### Example

  Before (collision):
  my-very-long-app-name-with-details-pvc-data-volume-producti.apps.example.com  ← PVC volume1
  my-very-long-app-name-with-details-pvc-data-volume-producti.apps.example.com  ← PVC volume2 (same!)

  After (unique):
  my-very-long-app-name-with-details-pvc-data-volum-a1b2c3d4.apps.example.com  ← PVC volume1
  my-very-long-app-name-with-details-pvc-data-volum-f9e8d7c6.apps.example.com  ← PVC volume2 (unique)

### DIFF
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -1037,13 +1037,15 @@ func getRouteHostName(client client.Client, namespacedName types.NamespacedName)
 	if len(routeNamePrefix) <= 62 {
 		return nil, nil
 	}
-	// if route prefix exceeds limits, a custom hostname will be provided
+	// if route prefix exceeds limits, truncate and append a hash suffix for uniqueness
 	ingressConfig := &configv1.Ingress{}
 	err := client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, ingressConfig)
 	if err != nil {
 		return nil, err
 	}
-	hostname := fmt.Sprintf("%s.%s", routeNamePrefix[:62], ingressConfig.Spec.Domain)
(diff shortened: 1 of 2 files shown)


aufi (approved):
LGTM

stillalearner:
Tested on OCP

PR #601, opened 2026-07-05
 :test_tube: validate : Automate test for verifying behavior with incorrect API file path

nachandr:
Fixes https://github.com/migtools/crane/issues/374

## Summary by CodeRabbit

* **Tests**
  * Added a new tier 1 end-to-end test for **MTA-863** to confirm `runner.Validate` returns an error when `APIResourcesFile` is invalid.
  * Covers two cases: a non-existent JSON file path and using a directory path instead of a file.
  * Verifies the error message includes OS-specific text for each failure type and cleans up the temporary working directory after the test.

PR #627, opened 2026-07-09
:test_tube: [release-0.10] validate : Automate test for verifying behavior with incorrect API file path

nachandr:
Cherry picking [fix](https://github.com/migtools/crane/pull/601)


stillalearner (approved):
lgtm

PR #630, opened 2026-07-12
:book: Update README with --run-as flag details

nachandr:
## Summary by CodeRabbit

* **Documentation**
  * Documented a new `--run-as` option for end-to-end test runs.
  * Clarified how admin vs. non-admin credentials are applied, including default behavior when the flag is omitted and what happens for invalid values.

### DIFF
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -142,6 +142,8 @@ Defined in `tests/tier0/e2e_suite_test.go` and `tests/tier1/e2e_suite_test.go`:
 - `--source-nonadmin-context` source kube context for namespace-admin (non-cluster-admin) user flows
 - `--target-nonadmin-context` target kube context for namespace-admin (non-cluster-admin) user flows
 - `--verbose-logs` enable command and output logging for framework runners
+- `--run-as` set to `admin` to run all tests with cluster-admin credentials. Omit this flag to
+run tests in their default mode (non-admin tests as non-admin, admin tests as admin).
 
 ## Adding a New Scenario
 


e2e-tests/README.md
```diff
@@ -142,6 +142,24 @@ Defined in `tests/tier0/e2e_suite_test.go` and `tests/tier1/e2e_suite_test.go`:
...
+
+This flag is validated in `config.ValidateAndLogRunAsFlag()` which is called in the `BeforeSuite` hooks. It accepts only two values:
+- `"admin"` - for cluster-admin mode
+- `""` (empty string) - for non-admin mode
+
+Any other value will cause the test suite to fail with an error.
+
+**Use cases:**
+- Run with `--run-as=admin` to test scenarios as cluster-admin and verify core functionality
+- Run without the flag (or `--run-as=""`) to test RBAC scenarios and validate that crane works with namespace-admin permissions
```
midays: The --run-as="" default documentation reads oddly, users don't pass an empty string, they just omit the flag. Documenting it as a valid value alongside "admin" makes it look like a two-option enum when it's really just a flag with a default. Suggested simplification:

```suggestion
- `--run-as=admin` Run all tests with cluster-admin credentials instead of the
  configured non-admin contexts. Omit this flag to run tests in their default
  mode (non-admin tests as non-admin, admin tests as admin).
```

This keeps the section to one bullet, no mention of empty string, and the default behavior is implied by "omit this flag". The ValidateAndLogRunAsFlag implementation detail and the two-value enum list can also be dropped from the docs, that's internal detail that doesn't help the reader.

midays (changes requested):

midays (approved):
LGTM

Issue #502, opened 2026-06-09
Test & Automate (tier 1): Migrate namespace with multiple secret types


midays:
## Summary
Add a test case to verify crane correctly migrates secrets of different
types (Opaque, TLS, dockerconfigjson) during a stateless migration.

## Polarion
https://polarion.engineering.redhat.com/polarion/#/project/MTAPathfinder/workitem?id=MTA-842

## Acceptance Criteria
- All secret types present on target after migration
- Secret data and type fields match source
- No errors during export/transform/apply pipeline

PR #635, opened 2026-07-12
:test_tube: Add e2e test for migrate namespace with multiple secret types

Tamar-Dinavetsky:
## Summary

  - Add e2e test for MTA-842: stateless migration of Kubernetes secrets with
  multiple types (Opaque, TLS, dockerconfigjson)
  - Verifies that crane export/transform/apply pipeline preserves all secret
  types intact
  - Validates secret type and data fields match source after migration to target
  cluster

 ## Test plan

  - [ ] Create 3 secrets (Opaque, TLS, dockerconfigjson) on source namespace
  - [ ] Run crane export/transform/apply pipeline - expect no errors
  - [ ] Verify all 3 secret manifests are present in output directory
  - [ ] Apply rendered manifests to target cluster
  - [ ] Verify each secret on target has correct type and non-empty data field

 ## Framework changes

  - Add VerifySecret() helper to test_helpers.go - fetches a secret by name,
  verifies type and data fields

  Polarion: [MTA-842](https://polarion.engineering.redhat.com/polarion/#/project/MTAPathfinder/workitem?id=MTA-842)
 [#502](https://github.com/migtools/crane/issues/502)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of migrated Kubernetes secrets, including secret type and required data.
  * Added coverage for exporting, transforming, and applying Opaque, TLS, and Docker registry secrets.
  * Confirmed migrated secrets retain their expected types in the target cluster.

### DIFF
--- a/e2e-tests/framework/test_helpers.go
+++ b/e2e-tests/framework/test_helpers.go
@@ -1,6 +1,7 @@
 package framework
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"regexp"
@@ -83,5 +84,35 @@ func ValidateClusterRBAC(kubectl KubectlRunner, bindings []ExpectedClusterRoleBi
 		}
 		log.Printf("CRB %s -> CR %s (subject: %s) verified", b.ClusterRoleBindingName, b.ClusterRoleName, b.SubjectName)
 	}
+	return nil
+}
+
+// Fetches a secret by name from the given namespace and cluster, parses the JSON response, and verifies the type
+// field matches the expected value. Returns a descriptive error if the secret cannot be fetched, the JSON cannot
+// be parsed, or the type does not match.
+func VerifySecret(kubectl KubectlRunner, namespace, secretName, expectedType string) error {
+	secretJson, err := kubectl.Run("get", "secret", secretName, "-n", namespace, "-o", "json")
+	if err != nil {
+		return fmt.Errorf("failed to get secret %s: %w", secretName, err)
+	}
+
+	var secretObj map[string]any
+	if err := json.Unmarshal([]byte(secretJson), &secretObj); err != nil {
+		return fmt.Errorf("failed to parse secret %s JSON: %w", secretName, err)
+	}
+
+	actualType, ok := secretObj["type"].(string)
+	if !ok {
+		return fmt.Errorf("secret %s has no type field", secretName)
+	}
+	if actualType != expectedType {
+		return fmt.Errorf("secret %s: expected type %q but got %q", secretName, expectedType, actualType)
+	}
+	data, ok := secretObj["data"].(map[string]any)
+	if !ok || len(data) == 0 {
+		return fmt.Errorf("secret %s has missing or empty data field", secretName)
+	}
+
+	log.Printf("Secret verified: name=%s type=%s\n", secretName, actualType)
 	return nil
 }
--- a/e2e-tests/framework/test_helpers.go
+++ b/e2e-tests/framework/test_helpers.go
@@ -87,20 +87,19 @@ func ValidateClusterRBAC(kubectl KubectlRunner, bindings []ExpectedClusterRoleBi
 	return nil
 }
 
-// Fetches a secret by name from the given namespace and cluster, parses the JSON response, and verifies the type
-// field matches the expected value. Returns a descriptive error if the secret cannot be fetched, the JSON cannot
-// be parsed, or the type does not match.
-func VerifySecret(kubectl KubectlRunner, namespace, secretName, expectedType string) error {
+// VerifySecret fetches a secret by name from the given namespace and cluster, parses the JSON response,
+// and verifies the type field matches the expected value and, when expectedData is non-nil, that every
+// key's base64-encoded value matches exactly. Returns a descriptive error if the secret cannot be fetched,
+// the JSON cannot be parsed, the type does not match, or any data key is missing/mismatched.
+func VerifySecret(kubectl KubectlRunner, namespace, secretName, expectedType string, expectedData map[string]string) error {
 	secretJson, err := kubectl.Run("get", "secret", secretName, "-n", namespace, "-o", "json")
 	if err != nil {
 		return fmt.Errorf("failed to get secret %s: %w", secretName, err)
 	}
-
 	var secretObj map[string]any
 	if err := json.Unmarshal([]byte(secretJson), &secretObj); err != nil {
 		return fmt.Errorf("failed to parse secret %s JSON: %w", secretName, err)
 	}
-
 	actualType, ok := secretObj["type"].(string)
 	if !ok {
 		return fmt.Errorf("secret %s/%s: type field is not a string, got %T", namespace, secretName, secretObj["type"])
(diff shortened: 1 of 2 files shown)


Tamar-Dinavetsky:
Tested on OCP successfully.

istein1:
1. MEDIUM: gofmt violation — spaces instead of tabs in VerifySecret. Two return statements in e2e-tests/framework/test_helpers.go use 6 spaces instead of a tab. All
  other lines use tabs correctly. Will likely fail CI lint.
  2. LOW: Inconsistent Expect pattern for multi-return functions. Uses Expect(kubectlSrc.Run(...)).Error().NotTo(HaveOccurred()) while the established codebase pattern
  captures return values explicitly (_, err = ...; Expect(err)...). Both work, but mixing patterns hurts readability.
  3. LOW: No availability guard for openssl binary. The test calls exec.Command("openssl", ...) directly. If the CI image lacks it, the test fails with a generic exec error. Adding exec.LookPath("openssl") with a Skip() would be more robust.
  4. TRIVIAL: Misleading DeferCleanup By() descriptions. Second and third By() blocks say "Cleanup source namespace and temp dir" / "Cleanup target namespace and temp 
  dir", but temp dir removal already happened in the first block.

Tamar-Dinavetsky:
@istein1 fixed , thanks for the review.

e2e-tests/framework/test_helpers.go
```diff
@@ -85,3 +86,33 @@ func ValidateClusterRBAC(kubectl KubectlRunner, bindings []ExpectedClusterRoleBi
...
+		return fmt.Errorf("secret %s/%s: expected type %q but got %q", namespace, secretName, expectedType, actualType)
+	}
+	data, ok := secretObj["data"].(map[string]any)
+	if !ok || len(data) == 0 {
+		return fmt.Errorf("secret %s/%s: data field is missing or empty, got %T", namespace, secretName, secretObj["data"])
+	}
+
+	log.Printf("Secret verified: name=%s type=%s\n", secretName, actualType)
+	return nil
+}
```
midays: Nice helper. One gap: it only checks `data` is non-empty, not that the content actually matches the source secret, a base64 re-encoding or truncation bug would slip through silently. The PVC tests in this suite (MTA-806/807) do full checksum comparison for the same reason; suggest doing the same here since secrets are even cheaper to compare exactly.

```suggestion
// Fetches a secret by name from the given namespace and cluster, parses the JSON response, and verifies the type
// field matches the expected value and, when expectedData is non-nil, that every key's base64-encoded value
// matches exactly. Returns a descriptive error if the secret cannot be fetched, the JSON cannot be parsed, the
// type does not match, or any data key is missing/mismatched.
func VerifySecret(kubectl KubectlRunner, namespace, secretName, expectedType string, expectedData map[string]string) error {
	secretJson, err := kubectl.Run("get", "secret", secretName, "-n", namespace, "-o", "json")
	if err != nil {
		return fmt.Errorf("failed to get secret %s: %w", secretName, err)
	}

	var secretObj map[string]any
	if err := json.Unmarshal([]byte(secretJson), &secretObj); err != nil {
		return fmt.Errorf("failed to parse secret %s JSON: %w", secretName, err)
	}

	actualType, ok := secretObj["type"].(string)
	if !ok {
		return fmt.Errorf("secret %s/%s: type field is not a string, got %T", namespace, secretName, secretObj["type"])
	}
	if actualType != expectedType {
		return fmt.Errorf("secret %s/%s: expected type %q but got %q", namespace, secretName, expectedType, actualType)
	}
	data, ok := secretObj["data"].(map[string]any)
	if !ok || len(data) == 0 {
		return fmt.Errorf("secret %s/%s: data field is missing or empty, got %T", namespace, secretName, secretObj["data"])
	}

	for key, expectedValue := range expectedData {
		actualValue, ok := data[key].(string)
		if !ok {
			return fmt.Errorf("secret %s/%s: expected data key %q not found or not a string", namespace, secretName, key)
		}
		if actualValue != expectedValue {
			return fmt.Errorf("secret %s/%s: data key %q does not match source value", namespace, secretName, key)
		}
	}

	log.Printf("Secret verified: name=%s type=%s\n", secretName, actualType)
	return nil
}

// GetSecretData fetches a secret by name and returns its data map with values left base64-encoded,
// exactly as returned by the Kubernetes API, suitable for direct comparison via VerifySecret's
// expectedData parameter without needing to decode/re-encode.
func GetSecretData(kubectl KubectlRunner, namespace, secretName string) (map[string]string, error) {
	secretJson, err := kubectl.Run("get", "secret", secretName, "-n", namespace, "-o", "json")
	if err != nil {
		return nil, fmt.Errorf("failed to get secret %s: %w", secretName, err)
	}
	var secretObj struct {
		Data map[string]string `json:"data"`
	}
	if err := json.Unmarshal([]byte(secretJson), &secretObj); err != nil {
		return nil, fmt.Errorf("failed to parse secret %s JSON: %w", secretName, err)
	}
	return secretObj.Data, nil
}
```

e2e-tests/tests/tier1/mta_842_secrets_migration_test.go
```diff
@@ -0,0 +1,110 @@
...
+		Expect(err).NotTo(HaveOccurred(), "openssl output: %s", out)
+
+		By("Create TLS secret on source")
+		_, err = kubectlSrc.Run("create", "secret", "tls", "tls-secret", "--cert="+certFile, "--key="+keyFile, "-n", namespace)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Create docker-registry secret on source")
+		_, err = kubectlSrc.Run("create", "secret", "docker-registry", "docker-secret", "--docker-server=quay.io", "--docker-username=user", "--docker-password=pass", "-n", namespace)
+		Expect(err).NotTo(HaveOccurred())
+
```
midays, line 78: Capture each secret's source data before the pipeline runs, so it can be compared exactly on target below.

```suggestion

		By("Capture source secret data for exact post-migration comparison")
		opaqueData, err := GetSecretData(kubectlSrc, namespace, "opaque-secret")
		Expect(err).NotTo(HaveOccurred())
		tlsData, err := GetSecretData(kubectlSrc, namespace, "tls-secret")
		Expect(err).NotTo(HaveOccurred())
		dockerData, err := GetSecretData(kubectlSrc, namespace, "docker-secret")
		Expect(err).NotTo(HaveOccurred())

```

e2e-tests/tests/tier1/mta_842_secrets_migration_test.go
```diff
@@ -0,0 +1,110 @@
...
+			Expect(matches).NotTo(BeEmpty(), "expected manifest for secret "+secretName+" in output directory")
+			log.Printf("Secret manifest found in output: %s\n", matches)
+		}
+
+		By("Apply rendered manifests to target")
+		log.Printf("Applying rendered manifests on target namespace %s from %s\n", namespace, paths.OutputDir)
+		Expect(ApplyOutputToTarget(kubectlTgt, namespace, paths.OutputDir)).NotTo(HaveOccurred())
+
+		By("Verify Opaque secret is present on target with correct type")
+		Expect(VerifySecret(kubectlTgt, namespace, "opaque-secret", "Opaque")).NotTo(HaveOccurred())
```
midays:
```suggestion
		Expect(VerifySecret(kubectlTgt, namespace, "opaque-secret", "Opaque", opaqueData)).NotTo(HaveOccurred())
```

e2e-tests/tests/tier1/mta_842_secrets_migration_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+		By("Verify tls secret is present on target with correct type")
+		Expect(VerifySecret(kubectlTgt, namespace, "tls-secret", "kubernetes.io/tls")).NotTo(HaveOccurred())
```
midays:
```suggestion
		Expect(VerifySecret(kubectlTgt, namespace, "tls-secret", "kubernetes.io/tls", tlsData)).NotTo(HaveOccurred())
```

e2e-tests/tests/tier1/mta_842_secrets_migration_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+		By("Verify docker secret is present on target with correct type")
+		Expect(VerifySecret(kubectlTgt, namespace, "docker-secret", "kubernetes.io/dockerconfigjson")).NotTo(HaveOccurred())
```
midays:
```suggestion
		Expect(VerifySecret(kubectlTgt, namespace, "docker-secret", "kubernetes.io/dockerconfigjson", dockerData)).NotTo(HaveOccurred())
```

e2e-tests/tests/tier1/mta_842_secrets_migration_test.go
```diff
@@ -0,0 +1,110 @@
...
+		out, err := cmd.CombinedOutput()
+		Expect(err).NotTo(HaveOccurred(), "openssl output: %s", out)
+
+		By("Create TLS secret on source")
+		_, err = kubectlSrc.Run("create", "secret", "tls", "tls-secret", "--cert="+certFile, "--key="+keyFile, "-n", namespace)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Create docker-registry secret on source")
+		_, err = kubectlSrc.Run("create", "secret", "docker-registry", "docker-secret", "--docker-server=quay.io", "--docker-username=user", "--docker-password=pass", "-n", namespace)
+		Expect(err).NotTo(HaveOccurred())
```
midays, line 77: While reviewing, dug into how crane actually handles ServiceAccount token secrets: it only whites out the **default** SA's token secret (via the `kubernetes.io/service-account.name: default` annotation, gated by `StripDefaultRBAC`), not by `.type == kubernetes.io/service-account-token`. So a token secret for a *custom* SA currently migrates as-is, token data included, which is stale/meaningless on the target. That behavior has zero test coverage today. Not asking to block this PR on it, but worth a follow-up case here or a separate test, e.g.:
```go
By("Verify default SA token secret is excluded from output")
matches, _ := filepath.Glob(filepath.Join(paths.OutputDir, "resources", namespace, "Secret__*default-token*"))
Expect(matches).To(BeEmpty(), "default SA token secret should be whited out")
```

midays (changes requested):

e2e-tests/tests/tier1/mta_842_secrets_migration_test.go
Tamar-Dinavetsky: @midays new test-> [#675](https://github.com/migtools/crane/issues/675)

midays (approved):
LGTM

PR #636, opened 2026-07-12
:test_tube:  [fix] mta_853 test logic so it will fit the test plan more accuratly

RanWurmbrand:
## Summary
- Add cluster-admin phase to MTA-853 test to complete the split-apply workflow per test plan NA-1
- Add `Overwrite` field to `ExportOptions` and `ApplyOptions` so the pipeline can re-run with `--overwrite`
- Create ServiceAccount and bind it to ClusterRoleBinding to properly test RBAC discovery

## Why
The previous MTA-853 test only ran the namespace-admin phase — it verified that cluster resources failed for non-admin and applied namespace resources, but never completed the workflow by having a cluster-admin apply the `_cluster/` resources. This didn't match test plan NA-1 which requires both phases with Pod Ready at the end.

## Test plan
- [x] Namespace-admin phase: export/transform/apply, verify cluster resources fail, apply namespace resources
- [x] Cluster-admin phase: re-run pipeline with --overwrite, verify cluster resources in `_cluster/`, apply them
- [x] Pod Ready after both phases

## Summary by CodeRabbit

* **Tests**
  * Updated the split-apply e2e test to run the export/transform/apply pipeline twice: once with namespace-admin permissions and again with cluster-admin permissions, with dedicated ServiceAccount creation and expanded RBAC cleanup.
  * Added overwrite capability to the e2e export/apply pipeline helpers so the pipeline can be re-run across phases, including assertions that namespace-scoped vs cluster-scoped outputs match expectations.

### DIFF
--- a/e2e-tests/tests/tier0/mta_853_split_apply_test.go
+++ b/e2e-tests/tests/tier0/mta_853_split_apply_test.go
@@ -120,13 +120,17 @@ var _ = Describe("Namespace-admin cluster-level migration", func() {
 		By("Cluster admin phase: Verifying cluster resources in output _cluster directory after cluster Admin phase")
 		Expect(utils.AssertKindsInOutput(paths.OutputDir, []string{"ClusterRole", "ClusterRoleBinding"})).NotTo(HaveOccurred())
 
-		By("Cluster admin phase: Applying namespace resources to target as namespace-admin")
+		By("Cluster admin phase: Applying cluster resources to target")
 		Expect(kubectlTgt.ApplyDir(filepath.Join(paths.OutputDir, "resources", "_cluster"))).NotTo(HaveOccurred())
 
 		By("Scaling target deployment and validating app")
 		Expect(kubectlTgtNonAdmin.ScaleDeployment(namespace, appName, 1)).NotTo(HaveOccurred())
 		Eventually(tgtAppNonAdmin.Validate, "5m", "10s").Should(Succeed())
 
+		By("Verifying ClusterRoleBinding on target references correct ClusterRole and ServiceAccount")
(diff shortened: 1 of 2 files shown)


aufi (approved):
LGTM

istein1:
Two minor findings:

  1. Misleading By() message (cosmetic) — The cluster-admin apply step says "Applying namespace resources to target as namespace-admin" but actually applies _cluster
  resources as cluster-admin. Should read "Applying cluster resources to target as cluster-admin".
  2. Missing explicit RBAC validation (suggestion) — MTA-852 uses ValidateClusterRBAC after apply; MTA-853 only checks pod readiness. Adding ValidateClusterRBAC would be
  more consistent and robust.

e2e-tests/tests/tier0/mta_853_split_apply_test.go
```diff
@@ -81,27 +83,46 @@ var _ = Describe("Namespace-admin cluster-level migration", func() {
...
+		//we reuse the same setup so we need to override for the second pipeline run
+		exportOpts.Overwrite = true
+		transformOpts.Overwrite = true
+		applyOpts.Overwrite = true
+		Expect(RunCranePipelineWithChecks(adminRunner, exportOpts, transformOpts, applyOpts)).NotTo(HaveOccurred())
+
+		By("Cluster admin phase: Verifying cluster resources in output _cluster directory after cluster Admin phase")
+		Expect(utils.AssertKindsInOutput(paths.OutputDir, []string{"ClusterRole", "ClusterRoleBinding"})).NotTo(HaveOccurred())
+
+		By("Cluster admin phase: Applying namespace resources to target as namespace-admin")
```
stillalearner: this statement looks incorrect

RanWurmbrand:
@coderabbitai review

PR #638, opened 2026-07-13
:test_tube: stateless tests to rel 0.10 branch

msajidmansoori12:
Cherry picking new stateless tests from main branch.


istein1 (approved):
Approving the cherry pick itself.

Code review notes for the tests by claude:

@nachandr ,
  1. MTA-864 stdout parser is fragile (mta_864_...test.go:81-109): The parseValidateRows function splits CLI table output by whitespace and distinguishes namespaced vs
  cluster-scoped rows by field count (5 vs 4). If the table format ever changes (e.g., a new column, or a resource name with spaces), this parser breaks silently. The
  test already validates the JSON report below — the stdout parsing adds coverage of the CLI display but is brittle. Consider validating only the structured JSON report
  and spot-checking stdout with ContainSubstring for key rows.
  2. MTA-864 uses a map key that loses uniqueness (mta_864_...test.go:99): rows[row.apiVersion+"|"+row.kind] would silently overwrite if two resources share the same
  apiVersion+kind (e.g., two Deployments from different namespaces). The current fixture has unique kinds so this doesn't cause a failure today, but it wouldn't catch a
  regression where a namespace's resources are dropped.
  3. MTA-861 offlineValidateFixture helper (mta_861_...test.go:24-61): Well-structured local helper. Other tier1 validate tests (MTA-863, MTA-850) do similar setup
  ad-hoc with filepath.Abs("../../testdata/...") which is working-directory-dependent and less robust. Not a problem for this PR but worth noting as a future cleanup
  opportunity on main.

@msajidmansoori12 ,
  4. MTA-839 cleanup ordering (mta_839_...test.go:38-46): The namespace DeferCleanup is registered before DeferCleanup(cleanup) from SetupActiveKubectlRunners. Since
  Ginkgo runs DeferCleanup in LIFO order, the RBAC cleanup runs first, then namespace deletion. This is the correct order — the RBAC teardown should happen before the
  namespace is deleted.

Issue #501, opened 2026-06-09
Test & Automate (tier 1): Migrate namespace with broken ConfigMap reference


midays:
## Summary
Add a test case to verify crane behavior when migrating a namespace
where a deployment references a non-existent ConfigMap. The expectation
is that crane migrates the resources as-is without failing, and the
broken reference lands on target unchanged.

## Polarion
https://polarion.engineering.redhat.com/polarion/#/project/MTAPathfinder/workitem?id=MTA-841

## Acceptance Criteria
- crane export/transform/apply completes without errors
- Deployment lands on target with the broken ConfigMap reference intact
- No silent data loss or modification of the broken reference

PR #654, opened 2026-07-14
:test_tube: Add e2e test for Deployment with missing ConfigMap reference

Tamar-Dinavetsky:
[#501](https://github.com/migtools/crane/issues/501)

## Summary
  Adds a tier1 E2E test to verify that Crane correctly migrates a namespace
  containing a Deployment that references a non-existent ConfigMap.

  #### How it works
  - Creates a source namespace with a Deployment that references `missing-configmap` via `envFrom` - a ConfigMap that does not exist on the
  cluster
  - Runs the full crane `export` → `transform` → `apply` pipeline
  - Verifies the pipeline completes without errors
  - Verifies the broken reference is preserved in the output YAML (no silent modification by transform plugins)
  - Applies the output to the target cluster and verifies the Deployment landed with the same broken reference intact

  #### Verification Logic
  - **Pipeline:** `RunCranePipelineWithChecks` - fails if any stage errors or produces empty output
  - **Output YAML:** `os.ReadFile` + `ContainSubstring` - ensures the broken ConfigMap reference was not dropped or modified during transform
  - **Target cluster:** `kubectl get deployment -o jsonpath= {.spec.template.spec.containers[0].envFrom[0].configMapRef.name}` - verifies the reference landed unchanged on target

## Summary by CodeRabbit

* **Tests**
  * Added new end-to-end coverage for a Deployment that references a missing ConfigMap.
  * Verified that migration keeps the broken/missing ConfigMap reference in the rendered output and after applying to the target cluster.
  * Added supporting Kubernetes test manifest and ensured temporary namespaces/resources are cleaned up.

### DIFF
--- a/e2e-tests/testdata/broken-ref-deployment.yaml
+++ b/e2e-tests/testdata/broken-ref-deployment.yaml
@@ -1,0 +1,21 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: broken-ref-deploy
+  namespace: mta-841-missing-cm
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: broken-ref-deploy
+  template:
+    metadata:
+      labels:
+        app: broken-ref-deploy
+    spec:
+      containers:
+        - name: nginx
+          image: nginx:alpine
+          envFrom:
+            - configMapRef:
+                name: missing-configmap
--- a/e2e-tests/testdata/broken-ref-deployment.yaml
+++ b/e2e-tests/testdata/broken-ref-deployment.yaml
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: broken-ref-deploy
-  namespace: mta-841-missing-cm
+  namespace: __NAMESPACE__
 spec:
   replicas: 1
   selector:
--- a/e2e-tests/tests/tier1/mta_841_missing_configmap_ref_test.go
+++ b/e2e-tests/tests/tier1/mta_841_missing_configmap_ref_test.go
@@ -1,0 +1,93 @@
+package e2e
+
+import (
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/konveyor/crane/e2e-tests/config"
+	. "github.com/konveyor/crane/e2e-tests/framework"
+	"github.com/konveyor/crane/e2e-tests/utils"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Missing ConfigMap reference migration", func() {
+	It("[MTA-841] crane migrates deployment with missing ConfigMap reference intact", Label("tier1"), func() {
+		namespace := "mta-841-missing-cm"
+		missingCMName := "missing-configmap"
+		scenario := NewMigrationScenario(
+			namespace,
+			namespace,
+			config.K8sDeployBin,
+			config.CraneBin,
+			config.SourceContext,
+			config.TargetContext,
+		)
+
+		kubectlSrc := scenario.KubectlSrc
+		kubectlTgt := scenario.KubectlTgt
+
+		paths, err := NewScenarioPaths("crane-export-*")
+		Expect(err).NotTo(HaveOccurred())
+		exportOpts := ExportOptions{Namespace: namespace, ExportDir: paths.ExportDir}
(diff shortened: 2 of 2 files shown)


Tamar-Dinavetsky:
Tested on OCP successfully.

RanWurmbrand:
lgtm, but i think we should wait for experienced  QE eyes before we merge

istein1:
1. MEDIUM: Missing export failure directory check. Established tests check failures/ dir after pipeline run. This test skips it. A resource could silently fail to
  export while the pipeline reports success — particularly relevant here since the test is specifically about a broken reference surviving the pipeline.
  
  2. MEDIUM: Expect() in DeferCleanup for temp dir removal can mask test failures. Namespace cleanup in the same block correctly uses if err + log.Printf, but temp dir
  cleanup uses Expect(os.RemoveAll(...)).To(Succeed()). If removal fails, the cleanup assertion overwrites the actual test failure. Other tests (e.g.,
  olm_whiteout_auditability_test.go) use the tolerant pattern for the same operation.
  
  3. LOW: Output YAML assertion uses string containment. ContainSubstring(missingCMName) could match a comment or annotation. The target-side check is more precise using
  jsonpath. Minor in practice since "missing-configmap" is unlikely to appear spuriously.
  
  4. LOW: Missing dry-run validation before applying to target. Some established tests include ValidateApplyDir (dry-run) before the real apply. Not universal, but a
  nice-to-have.

Tamar-Dinavetsky:
@istein1  @RanWurmbrand  fixed , thanks for the review.

e2e-tests/testdata/broken-ref-deployment.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: broken-ref-deploy
+  namespace: mta-841-missing-cm
```
midays: The namespace is hardcoded here and duplicated as a separate string literal in the test (`namespace := "mta-841-missing-cm"`). The repo already has a convention for this,`__NAMESPACE__` placeholder + `strings.ReplaceAll`, used in `rolebinding_namespace_admin.yaml` and the OLM testdata files (see `GrantNamespaceAdminToUser` in `rbac.go`). Using it here avoids the two files silently drifting out of sync.

```suggestion
  namespace: __NAMESPACE__
```

e2e-tests/tests/tier1/mta_841_missing_configmap_ref_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+			By("Cleanup namespace and temp dir")
+			if err := os.RemoveAll(paths.TempDir); err != nil {
+				log.Printf("cleanup: failed to remove temp dir: %v", err)
+			}
+		})
+
+		By("Create source namespace")
+		_, err = kubectlSrc.Run("create", "namespace", namespace)
+		Expect(err).NotTo(HaveOccurred())
```
midays: This errors loudly on `AlreadyExists`, unlike the target-side creation later in this same test which uses the tolerant `CreateNamespace` helper. Worth using the same helper here for consistency and to avoid a spurious failure if a prior run's cleanup didn't fully complete.

```suggestion
		Expect(kubectlSrc.CreateNamespace(namespace)).NotTo(HaveOccurred())
```

e2e-tests/tests/tier1/mta_841_missing_configmap_ref_test.go
```diff
@@ -0,0 +1,105 @@
...
+		log.Printf("Created source namespace %s\n", namespace)
+
+		By("Apply Deployment with missing ConfigMap reference to source")
+		deploymentYAML, err := utils.ReadTestdataFile("broken-ref-deployment.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		_, err = kubectlSrc.RunWithStdin(deploymentYAML, "apply", "-f", "-")
```
midays, line 60: Pairs with the `__NAMESPACE__` placeholder suggestion on the testdata file, substitute it in before applying.

```suggestion
		deploymentYAML, err := utils.ReadTestdataFile("broken-ref-deployment.yaml")
		Expect(err).NotTo(HaveOccurred())
		deploymentYAML = strings.ReplaceAll(deploymentYAML, "__NAMESPACE__", namespace)
		_, err = kubectlSrc.RunWithStdin(deploymentYAML, "apply", "-f", "-")
```

midays (changes requested):
LGTM, few comments

midays (approved):
LGTM

PR #662, opened 2026-07-15
:seedling: Enable build release binaries for release-* branches

### LINKED ISSUE #652: Enable build-release-binaries.yml for release-0.10.
Try to extract reusable logic in workflows then enable it for release-0.10 branch.

msajidmansoori12:
## Summary by CodeRabbit

* **New Features**
  * Added automated nightly builds for the release-0.10 line.
  * Expanded release build coverage to include multiple Linux, macOS, and Windows architectures.
  * Release artifacts now include version metadata and SHA256 checksum files.

* **Improvements**
  * Standardized release binary creation through a reusable build process.
  * Added validation to confirm version information is embedded in generated binaries.

### DIFF
--- a/.github/workflows/build-release-binaries.yml
+++ b/.github/workflows/build-release-binaries.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - release-0.*
     tags:
       - "*"
   schedule:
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-build-release-binaries.yml
```diff
@@ -0,0 +1,99 @@
...
+          - goos: darwin
+            goarch: arm64
+          - goos: windows
+            goarch: amd64
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.checkout_ref }}
```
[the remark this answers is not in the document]
msajidmansoori12: It's fine to have this.

.github/workflows/build-release-binaries.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - release-0.*
```
aufi, line 7: A small suggestion on having generally `release-*` (instead of require version starting with `0.`).

aufi (approved):
Thanks for the PR, LGTM.

.github/workflows/build-release-binaries.yml
msajidmansoori12: Hi @aufi , thanks for the suggestion, i actually deliberately kept it to only run release-0.* branches , because I was checking the branches on crane and I see a lot of release-* branches which I am not sure are relevant or not.
release-1.7.5 , relase-1.7.6 and so on.  If you think we should enable it anyway then i'll do it.

aufi: ah, good point with obsolete branches, then let's keep change you proposed. We might discuss what to do with those branches and update later.

PR #671, opened 2026-07-17
:bug: fix-sanitize plugin manifest name in plugin-manager add to prevent p…

### LINKED ISSUE #670: [BUG] Unsanitized plugin name in `plugin-manager add` allows plugin override and arbitrary file write
### Summary

The `plugin-manager add` command uses the `name` field from the plugin manifest directly as a filename when writing the downloaded binary to disk, without any sanitization or validation. This enables two attack vectors:

1. **Plugin Override** — A malicious plugin repository can set the manifest `name` to match an existing plugin (e.g. `OpenShiftPlugin`), silently overwriting it with a trojanized version. The trojanized plugin can perform all legitimate transformations while injecting rogue Kubernetes resources into the transform output, which the user then unknowingly applies to their cluster.

2. **Path Traversal** — The manifest `name` can contain `../../` sequences, allowing the attacker to write an executable file (with 0777 permissions) to any location the user has write access to. This is significantly worse if crane is run with `sudo`, as it could overwrite system binaries.

### Root Cause

In [`cmd/plugin-manager/add/add.go` line 232](https://github.com/migtools/crane/blob/main/cmd/plugin-manager/add/add.go#L232):

```go
pluginBinary, err := os.OpenFile(filepath+"/"+filename, syscall.O_RDWR|syscall.O_CREAT|syscall.O_TRUNC, 0777)
```

The `filename` parameter comes from `value.Name` in the plugin manifest YAML. There is no validation that:
- The name matches what the user actually requested to install
- The name does not contain path traversal characters (`/`, `\`, `..`)

Additionally, the `Validate` step in `add.go` only checks if the user-provided name (e.g. `CoolPlugin`) already exists in the plugins directory — it does not check the actual `name` from the manifest (e.g. `OpenShiftPlugin`), so the overwrite is not detected.

### Proof of Concept

A full end-to-end PoC is available at: https://github.com/RanWurmbrand/compromised-plugins

RanWurmbrand:
closes: #670 

## Summary

The `plugin-manager add` command uses the `name` field from the plugin manifest as the binary filename without any sanitization. This allows a malicious plugin repository to:

1. **Override existing plugins** — a manifest can set its `name` to an already-installed plugin (e.g. `OpenShiftPlugin`), silently replacing it with a trojanized version
2. **Write files anywhere on disk** — a manifest `name` containing `../../` sequences escapes the plugins directory, writing an executable (0777 permissions) to arbitrary paths

### Root Cause

In `downloadBinary()`, `value.Name` from the manifest YAML is passed directly to `os.OpenFile` as the filename with no validation.

### Fix

Added a `validateFileInput()` function that runs before any download:
- **Rejects** names containing `/`, `\`, or `..` — blocks path traversal
- **Rejects** names that match an already-installed plugin binary — blocks silent override

### PoC

Full end-to-end proof of concept: https://github.com/RanWurmbrand/compromised-plugins

## Test plan

- [ ] `crane plugin-manager add EvilPlugin` with a manifest name containing `../../` is rejected with an error
- [ ] `crane plugin-manager add CoolPlugin` with a manifest name matching an existing plugin (`OpenShiftPlugin`) is rejected with an error
- [ ] Normal plugin installation (manifest name matches requested name, no traversal) still works

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for plugin filenames to reject unsafe paths and prevent path traversal.
  * Prevented downloads from overwriting an existing plugin binary.
  * Invalid requests now fail before downloading or writing files.

### DIFF
--- a/cmd/plugin-manager/add/add.go
+++ b/cmd/plugin-manager/add/add.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"syscall"
 
 	"github.com/konveyor/crane/internal/flags"
@@ -201,7 +202,23 @@ func (o *Options) run(args []string) error {
 	return nil
 }
 
+func validateFileInput(filepath, filename string) error {
+	if strings.Contains(filename, "/") || strings.Contains(filename, "\\") || strings.Contains(filename, "..") {
+		return fmt.Errorf("invalid plugin name %q: must not contain path separators or traversal sequences", filename)
+	}
+
+	existingPath := filepath + "/" + filename
+	if _, err := os.Stat(existingPath); err == nil {
+		return fmt.Errorf("a plugin named %q already exists at %s, please remove it first before installing", filename, existingPath)
+	}
+	return nil
+}
+
 func downloadBinary(filepath string, filename string, url string, log *logrus.Logger) error {
+	if err := validateFileInput(filepath, filename); err != nil {
+		return err
+	}
+	
 	var binaryContents io.Reader
 	isUrl, url := plugin.IsUrl(url)
 	if !isUrl {
--- a/cmd/plugin-manager/add/add.go
+++ b/cmd/plugin-manager/add/add.go
@@ -202,14 +202,16 @@ func (o *Options) run(args []string) error {
 	return nil
 }
 
-func validateFileInput(filepath, filename string) error {
-	if strings.Contains(filename, "/") || strings.Contains(filename, "\\") || strings.Contains(filename, "..") {
-		return fmt.Errorf("invalid plugin name %q: must not contain path separators or traversal sequences", filename)
+func validateFileInput(dir, filename string) error {
+	if filename == "" || filename == "." || filename == ".." || strings.ContainsRune(filename, '\\') || filename != filepath.Base(filename) {
+		return fmt.Errorf("invalid plugin name %q: must be a bare file name without path separators or traversal sequences", filename)
 	}
 
-	existingPath := filepath + "/" + filename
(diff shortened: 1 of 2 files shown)


cmd/plugin-manager/add/add.go
```diff
@@ -201,7 +202,23 @@ func (o *Options) run(args []string) error {
 	return nil
 }
 
+func validateFileInput(filepath, filename string) error {
```
aufi: It is better don't the use filepath when it is already a package name in this file.

cmd/plugin-manager/add/add.go
```diff
@@ -201,7 +202,23 @@ func (o *Options) run(args []string) error {
...
+	if strings.Contains(filename, "/") || strings.Contains(filename, "\\") || strings.Contains(filename, "..") {
```
aufi: The validation needs some hardening (consider using filepath.Base pattern instead of blocklist for the filename) and ans please include unit tests.

aufi (commented):
Thanks for finding this problem and creating PR! Left few comments.

RanWurmbrand:
@aufi let me know if these changes is what you had in mind

aufi (approved):
@RanWurmbrand The PR looks good to me! Would you update branch&merge or should I do it?

RanWurmbrand:
@aufi ill do it by EOD, thanks for suggesting!

RanWurmbrand:
@aufi looks like tammy merged a pr with fix's for this few days ago [#913](https://github.com/migtools/crane/pull/913l)

ill think it will be cleaner if ill just create new pr and modifed the latest, and add the unit tests, (there are cases which the current version miss)

alternativly i can close this one and we stay with what we have

PR #637, opened 2026-07-13
:bug: Fix 813 test

stillalearner:
## Summary by CodeRabbit

* **Tests**
  * Updated the PVC transfer test to use consistent application context values during transfer validation.
  * Removed an outdated note from the test scenario.

PR #745, opened 2026-08-02
:test_tube: mta_812: adopt --ordered apply now that crane#266 is fixed

### LINKED ISSUE #686: investigate tier0 tests with data transfer and assure the assertion works

midays:
## Summary
- `mta_812_role_migration_test.go` had a two-pass `kubectl apply` workaround for crane#266 (RoleBinding sorting alphabetically before its Role on a fresh namespace apply). #266 is now fixed via the opt-in `--ordered` apply flag, which prefixes output filenames with a dependency-aware order number (e.g. `300_Role_*`, `310_RoleBinding_*`).
- This test now sets `Ordered: true` and does a single apply pass instead of the old skip-dry-run-then-retry workaround.
- Updated the Role/RoleBinding manifest-presence glob assertions to match regardless of the ordering prefix (`*Role_*.yaml` / `*RoleBinding_*.yaml`), since the old exact-prefix globs would've silently stopped matching once ordering was enabled.
- Kept the `BUG #266` title/label tag on the test for historical traceability, per team convention, even though the workaround itself is gone.

Other `BUG #NNN`-tagged tests were audited as part of this pass:
- `mta_807`, `mta_811` (crane#213): already fully fixed upstream (PR #603) with no remaining workaround — no changes needed.
- `mta_813` (crane#330 fixed / crane#408 still open): crane#330's workaround was already removed (PR #637). The PVC-content-integrity assertion tied to crane#408 stays commented out — verified by temporarily re-enabling and running it locally, which reproduced the bug (crane's own output showed `Status: Failed`, `Files Sent: 0`, rsync exit code 12, yet `crane transfer-pvc` still reported `"PVC data copy: succeeded"` and exited 0). Reverted back to the commented-out state since crane#408 is not fixed yet.
- `mta_862`: no change needed — it's already the dedicated regression test for the `--ordered` fix.

## Test plan
- [x] `go build ./...` and `go vet ./...` pass in `e2e-tests/`
- [x] Ran `mta_812` locally against `src`/`tgt` minikube clusters — passes with a single ordered apply pass (verified ordered filenames `300_Role_*`/`310_RoleBinding_*` in output, single dry-run + apply, no first-pass/second-pass workaround)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment ordering so dependent resources are applied in the correct sequence.
  * Updated manifest validation to reliably accept correctly ordered resource filenames.
  * Simplified target application to use a single standard apply step, improving consistency.

### DIFF
--- a/e2e-tests/tests/tier0/mta_812_role_migration_test.go
+++ b/e2e-tests/tests/tier0/mta_812_role_migration_test.go
@@ -92,7 +92,8 @@ var _ = Describe("Role and RoleBinding migration", func() {
 		exportOpts := ExportOptions{Namespace: namespace, ExportDir: paths.ExportDir}
 		transformOpts := TransformOptions{ExportDir: paths.ExportDir, TransformDir: paths.TransformDir}
 		applyOpts := ApplyOptions{TransformDir: paths.TransformDir,
-			OutputDir: paths.OutputDir}
+			OutputDir: paths.OutputDir,
+			Ordered:   true}
 		DeferCleanup(func() {
 			By("Cleanup source and target resources")
 			if err := CleanupScenario(paths.TempDir, srcApp, tgtApp); err != nil {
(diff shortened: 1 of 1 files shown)


istein1:
The :test_tube: prefix suggests this is a test-only change, but the PR adds new user-facing warning functionality in addition to its tests. Could you update the title to use :sparkles:
 (or feat:) instead, to better reflect that this is a feature addition? Something like :sparkles: Warn when exporting image-related OCP resources (crane#681).

aufi (approved):
LGTM

My $0.02 on the gitmoji prefix, I'm not sure what is user-facing warning in test suite (E2E test suite is not even part of the crane binary), clear removal of second apply command call with its warning makes sense to me.

istein1:
Apologies, posted this comment on the wrong PR! This was intended for [#746](https://github.com/migtools/crane/pull/746).

stillalearner (approved):

Issue #289, opened 2026-04-20
[BUG] [transfer-pvc] restrictedContainers.ApplyTo() applies --omit-dir-times unconditionally


RanWurmbrand:
### restrictedContainers.ApplyTo() applies --omit-dir-times unconditionally

**What version of crane are you running, and what are your clusters+platform**

Latest main branch. Issue exists in `cmd/transfer-pvc/transfer-pvc.go` lines 793-801.

**What did you expect to happen?**

The `--omit-dir-times` flag should respect the boolean value like other options (Groups, Owners, DeviceFiles, SpecialFiles) which use `opts.X = bool(!r)`.

**What actually happened?**

`--omit-dir-times` is always appended regardless of the boolean value, while other flags check the boolean first.

Currently doesn't matter in practice since `restrictedContainers(true)` is hardcoded at line 452, but the inconsistency could cause issues if that changes.

**Please include any relevant logs or errors**

N/A — discovered via code review.

PR #748, opened 2026-08-03
:bug: fix: apply --omit-dir-times conditionally in restrictedContainers

Tamar-Dinavetsky:
https://github.com/migtools/crane/issues/289

## Summary
  - `--omit-dir-times` was unconditionally appended to rsync extras in `restrictedContainers.ApplyTo()`, unlike all
  - Fixed to only append when `r` is `true` (restricted mode on), consistent with `Groups`, `Owners`,
  `DeviceFiles`, and `SpecialFiles`
  
  ## Impact
  No functional change - `restrictedContainers(true)` is hardcoded at the call site.
  Prevents a silent bug if the call site ever passes `restrictedContainers(false)`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transfer operations now apply directory timestamp omission only when restricted containers are enabled, improving behavior in other configurations.


midays (approved):
LGTM

Tamar-Dinavetsky:
[review-docs]

PR #749, opened 2026-08-03
:test_tube: mta-875 add tier1 e2e test for skipping PV migration with pre-migrated PVC data

midays:
[Jenkins run](https://jenkins-csb-migrationqe-main.dno.corp.redhat.com/job/mta/job/crane-ocp-tests-runner/384/console)

## Tests

- Added end-to-end coverage for migrating applications with pre-populated persistent volumes.
- Verified volume data is transferred before application migration and remains intact afterward.
- Confirmed migrated applications reconnect to existing volumes and serve the expected content.
- Validated that generated migration output excludes unnecessary persistent volume resources when storage is already available.
- Added checks for transfer completion, data integrity, resource cleanup, and successful application redeployment.

### DIFF
--- a/e2e-tests/tests/tier1/mta_875_skip_premigrated_pvc_data_test.go
+++ b/e2e-tests/tests/tier1/mta_875_skip_premigrated_pvc_data_test.go
@@ -1,0 +1,154 @@
+package e2e
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/konveyor/crane/e2e-tests/config"
+	. "github.com/konveyor/crane/e2e-tests/framework"
+	"github.com/konveyor/crane/e2e-tests/utils"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func getPodName(k KubectlRunner, namespace, label string) (string, error) {
+	out, err := k.Run("get", "pods", "-n", namespace, "-l", "app="+label, "-o", "jsonpath={.items[0].metadata.name}")
+	if err != nil {
+		return "", err
+	}
+	name := strings.TrimSpace(out)
+	if name == "" {
+		return "", fmt.Errorf("no pod found in namespace %q with label app=%s", namespace, label)
+	}
+	return name, nil
+}
+
+func readPodFile(k KubectlRunner, namespace, pod, path string) (string, error) {
+	return k.Run("exec", pod, "-n", namespace, "--", "cat", path)
+}
+
+func writePodFile(k KubectlRunner, namespace, pod, path, content string) error {
+	_, err := k.RunWithStdin(content, "exec", pod, "-n", namespace, "-i", "--", "sh", "-c", "cat > "+path)
+	return err
+}
+
+func curlStatusCode(k KubectlRunner, namespace, pod string) (string, error) {
+	return k.Run("exec", pod, "-n", namespace, "--", "sh", "-c", "curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:8080/")
+}
+
+var _ = Describe("Skip PV migration when PV data was already migrated ahead of time", func() {
+	It("[MTA-875] Skip PV migration when PV data was already migrated ahead of time", Label("tier1", "pvc-transfer"), func() {
+		const podLabel = "nginx"
+		appName := "nginxpv"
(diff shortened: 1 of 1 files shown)


e2e-tests/tests/tier1/mta_875_skip_premigrated_pvc_data_test.go
```diff
@@ -0,0 +1,154 @@
...
+		By("Remove the temporary app, leaving the pre-populated PVCs bound")
+		_, err = kubectlTgt.Run("delete", "deployment", "nginx-deployment", "-n", tgtApp.Namespace, "--wait=true")
+		Expect(err).NotTo(HaveOccurred())
+		_, err = kubectlTgt.Run("delete", "service", "my-nginx", "-n", tgtApp.Namespace)
+		Expect(err).NotTo(HaveOccurred())
+		tgtPVCs, err := ListPVCs(tgtApp.Namespace, "", tgtApp.Context)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(tgtPVCs).To(HaveLen(2), "expected the two pre-populated PVCs to remain after removing the temporary app")
+
+		By("Run crane export/transform/apply pipeline (no transfer-pvc: PV migration is skipped)")
```
msajidmansoori12: @midays can you please recheck the polarion test case , the steps seem to be different with what the test is actually doing.

msajidmansoori12 (changes requested):

e2e-tests/tests/tier1/mta_875_skip_premigrated_pvc_data_test.go
```diff
@@ -0,0 +1,130 @@
...
+		By("Verify the rendered manifests contain no PVC/PV: the already-migrated PVC is not re-transferred or re-created")
+		outputFiles, err := utils.ListFilesRecursivelyAsList(paths.OutputDir)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(outputFiles).NotTo(BeEmpty())
+		for _, f := range outputFiles {
+			Expect(f).NotTo(ContainSubstring("PersistentVolume"), "output should not contain a PVC/PV manifest file: %s", f)
+			content, err := os.ReadFile(filepath.Join(paths.OutputDir, f))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(content)).NotTo(ContainSubstring("PersistentVolume"), "output file %s should not reference a PVC/PV", f)
+		}
```
msajidmansoori12, line 164: I think this assertion is still a bit too weak for what the step description says. We only scan for the substring PersistentVolume, so this can miss PersistentVolumeClaim manifests and also lower-case persistentVolumeClaim references inside workload YAML.

Could we switch this to a kind-based check instead, e.g. utils.AssertNoKindsInOutput(..., []string{"PersistentVolume", "PersistentVolumeClaim"})? That would make the intent much clearer and would better prove that the main migration path is not trying to recreate PVC/PV objects.

e2e-tests/tests/tier1/mta_875_skip_premigrated_pvc_data_test.go
```diff
@@ -0,0 +1,130 @@
...
+		}, "60s", "2s").ShouldNot(ContainSubstring("rsync"))
+		Eventually(func() (string, error) {
+			return kubectlTgt.Run("get", "pods", "-n", tgtApp.Namespace, "-o", "name")
+		}, "60s", "2s").ShouldNot(ContainSubstring("rsync"))
+
+		By("Verify the destination PVC exists")
+		tgtPVCs, err := ListPVCs(tgtApp.Namespace, "", tgtApp.Context)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(tgtPVCs).To(HaveLen(1), "expected the pre-migrated PVC to exist on target")
+
```
msajidmansoori12, line 114: This is much closer now that the test actually runs transfer-pvc. One thing I think is still missing from the testcase flow: here we only verify that the destination PVC exists, but not that its contents already match the source before export/transform/apply runs.

Would you mind validating the transferred data at this point as well, maybe by mounting the target PVC in a temporary verifier pod and checking the MD5 there? Right now the MD5 comparison only happens after the target app is deployed, so we never isolate whether transfer-pvc itself produced the correct target PVC state.

msajidmansoori12 (changes requested):

msajidmansoori12 (approved):
lgtm

Issue #778, opened 2026-08-09
improve log messages in transform directory.


Tamar-Dinavetsky:
parent: https://github.com/migtools/crane/issues/297

PR #780, opened 2026-08-09
:sparkles: feat(audit): improve logging in transform and export commands

Tamar-Dinavetsky:
https://github.com/migtools/crane/issues/778
Adds missing log statements across export and transform commands as a first step toward the persistent audit log feature.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic details for validation, plugin discovery, path resolution, and transformation-stage failures.
  * Error messages now provide clearer information for invalid selections, duplicate requests, ambiguous results, and plugin-loading issues.

* **Enhancements**
  * Added structured progress and completion logging across export and transformation workflows.
  * Improved visibility into discovered plugins, stage creation, reconciliation, cleanup, and resource resolution.
  * Logging now consistently respects configured verbosity levels and provides a reliable fallback when no logger is configured.


cmd/export/export.go
```diff
@@ -110,15 +124,18 @@ func (o *ExportOptions) Validate() error {
 			{"--token", o.configFlags.BearerToken},
 		} {
 			if f.val != nil && *f.val != "" {
+				log.Errorf("Cannot use --context with %s", f.flag)
 				return fmt.Errorf("cannot use --context with %s; it overrides the value defined in the context", f.flag)
 			}
 		}
 	}
 	if o.asExtras != "" && *o.configFlags.Impersonate == "" && len(*o.configFlags.ImpersonateGroup) == 0 {
+		log.Errorf("--as-extras requires specifying a user or group to impersonate")
```
stillalearner: should it be changed to Debug? 

as there is already Error being returned

stillalearner: and similarly other ones...wherever return is already there.

cmd/transform/transform.go
```diff
@@ -279,22 +309,30 @@ func (o *Options) run() error {
...
 		if len(existingStages) == 0 {
 			// No stages exist - load all available plugins and create stages for each
 			allPlugins, err := plugin.GetFilteredPlugins(pluginDir, o.SkipPlugins, log)
 			if err != nil {
+				log.Debugf("Failed to load plugins from %q: %v", pluginDir, err)
 				return fmt.Errorf("failed to load plugins: %w", err)
 			}
 
 			if len(allPlugins) == 0 {
+				log.Errorf("No plugins found in plugin directory %q", pluginDir)
```
stillalearner: here

Tamar-Dinavetsky: missed it, sorry.
 Updated

Tamar-Dinavetsky:
test on ocp- succeed

cmd/transform/transform.go
```diff
@@ -59,19 +59,28 @@ func (o *Options) Complete(c *cobra.Command, args []string) error {
 }
 
 func (o *Options) Validate() error {
+	log := logrus.StandardLogger()
+	if o.globalFlags != nil {
+		log = o.globalFlags.GetLogger()
```
stillalearner: in the globalFlag. GetLogger(), we need to get the same logger everytime if already exists. 

RIght now it creates new logger everytime

stillalearner: type GlobalFlags struct {
    ConfigFile string
    Debug      bool
    logger     *logrus.Logger
}

func (g *GlobalFlags) GetLogger() *logrus.Logger {
    if g.logger == nil {
        g.logger = logrus.New()
        if g.Debug {
            g.logger.SetLevel(logrus.DebugLevel)
        }
    }
    return g.logger
}


something like this is needed..

stillalearner: Then only we would be able to configure audit file from one single place

cmd/export/export.go
```diff
@@ -53,6 +53,10 @@ type ExportOptions struct {
 // Complete loads kubeconfig context, namespace, and parses --as-extras into o.extras.
 func (o *ExportOptions) Complete(c *cobra.Command, args []string) error {
 	var err error
+	log := logrus.StandardLogger()
```
stillalearner: I suggest adding a log flag in the Options struct for each command. 

log         *logrus.Logger

and then initialise it in Complete function for each command like this:

 o.log = o.globalFlags.GetLogger()

where GetLogger should also be updated to not create a new logger everytime as mentioned in my other comment

cmd/transform/listplugins/listplugins.go
```diff
@@ -32,6 +33,12 @@ type Flags struct {
 
 func (o *Options) Complete(c *cobra.Command, args []string) error {
 	// TODO: @sseago
+	if o.globalFlags != nil {
+		o.log = o.globalFlags.GetLogger()
+	}
+	if o.log == nil {
+		o.log = logrus.StandardLogger()
+	}
```
aufi: Conditional assignment of log looks to be repeated on multiple places, I'd suggest consider creating a method with this and call it from multiple commands.

cmd/transform/transform.go
```diff
@@ -261,16 +292,19 @@ func (o *Options) run() error {
 			}
 			log.Infof("Running stage: %s", stageName)
 			if err := o.runStageWithCleanup(orchestrator, selector, stageDir, !stageExists, log); err != nil {
+				log.Errorf("Failed to run stage %q: %v", stageName, err)
 				return err
 			}
 		}
+		log.Infof("Transform complete")
```
aufi: Not a hard requirement, but if possible, it would be great having Infof messages (within a single command (even different params)) unique? It might help with debugging on error reports.

(this instructions file transformation complete is same as line 320 CLI transformation complete, not a big deal, but considering modifying this to Transform by instructions file complete, or similar)

aufi (commented):
Overall looks good to me. Great additions to Error logging output, otherwise left few comments for consideration or further discussion.

internal/flags/global_flags.go
```diff
@@ -19,11 +20,15 @@ func (g *GlobalFlags) ApplyFlags(cmd *cobra.Command) {
 }
 
 func (g *GlobalFlags) GetLogger() *logrus.Logger {
-	log := logrus.New()
+	if g.logger == nil {
+		g.logger = logrus.New()
+	}
 	if g.Debug {
```
stillalearner: this check can be moved inside above block, so that it is called once only

stillalearner (approved):
LGTM

aufi (approved):
LGTM, nice work Tammy!

Tamar-Dinavetsky:
thanks @stillalearner  @aufi  for the reviews

Issue #779, opened 2026-08-09
improve log messages in apply directory


Tamar-Dinavetsky:
parent: https://github.com/migtools/crane/issues/297

PR #784, opened 2026-08-10
:sparkles: Feat/audit logging apply

Tamar-Dinavetsky:
https://github.com/migtools/crane/issues/779
Adds structured logging across apply commands as the first step toward the persistent audit log feature.

## Summary by CodeRabbit

* **Improvements**
  * Added detailed logging throughout the apply workflow, including startup, validation, path handling, stage discovery, and completion.
  * Improved visibility into transform-directory failures, missing stages, configuration issues, and apply errors.
  * Error reporting now preserves the underlying failure details for easier troubleshooting.
  * Apply progress and outcomes are now easier to follow in command output.
  * Existing apply behavior remains unchanged.


Tamar-Dinavetsky:
test on ocp- succeed

cmd/apply/apply.go
```diff
@@ -128,33 +135,41 @@ func addFlagsForOptions(o *Flags, cmd *cobra.Command) {
...
 
 	if _, err := os.Stat(outputDir); err == nil {
 		if !o.Overwrite {
+			log.Debugf("Output directory %q already exists; use --overwrite to replace it", outputDir)
 			return fmt.Errorf("output directory %q already exists; use --overwrite to replace it", outputDir)
 		}
 		if err := os.RemoveAll(outputDir); err != nil {
-			return fmt.Errorf("failed to clear output directory: %w", err)
+			log.Errorf("Failed to clear output directory %q: %v", outputDir, err)
+			return err
```
msajidmansoori12: Hi @Tamar-Dinavetsky ,
Thanks for the extra logging here — that part looks good.

One thing to restore: on the RemoveAll, MkdirAll, and DiscoverStages failure paths we used to wrap the error (failed to clear output directory: %w, etc.) and now we log.Errorf(...) and return err.

log.Errorf prints to stderr, but it isn’t what cobra shows as the command error. The user still gets cobra’s Error: line from the returned value, so they end up with something like:

```
ERRO[0000] Failed to clear output directory "/path/output": permission denied
Error: permission denied
```
The first line is useful; the second one lost the “what failed” context we used to have.

Could you keep the wrap as well as the log? Something like:

```
log.Errorf("Failed to clear output directory %q: %v", outputDir, err)
return fmt.Errorf("failed to clear output directory %q: %w", outputDir, err)
```
Same idea for creating the output dir and discovering stages. The invalid kustomize-args: %w return already does this.

cmd/apply/apply.go
```diff
@@ -26,6 +27,7 @@ type Options struct {
 	Flags
 	// Positional arguments for stage selection
 	RequestedStages []string
+	log             *logrus.Logger
```
msajidmansoori12, line 30: @Tamar-Dinavetsky 
`o.log` is set in `Complete` (line 48) but never read. `Validate `and run both call `GetLoggerOrDefault() `again instead.

If nothing uses the field, we can drop it and the extra `logrus` import. Export actually uses `o.log;` here it looks leftover from that pattern.

Happy to keep it if you plan to use it in a follow-up — just unused as written.

msajidmansoori12 (changes requested):

cmd/apply/apply.go
Tamar-Dinavetsky: Hi @msajidmansoori12,

  Thanks for the review! I wanted to check before making changes - looking at export.go and transform.go, both
  follow this pattern for system/infrastructure errors:

  ```
log.Errorf("Error clearing export directory: %v", err)
  return err  // raw, no wrapping

```
  I followed the same pattern in apply.go to stay consistent.

  For user-facing validation errors (bad input, invalid flags, etc.) we do return a descriptive fmt.Errorf  e.g.
 ` return fmt.Errorf("invalid kustomize-args: %w", err)` - so those already have full context in the cobra output.

  For system errors (filesystem ops, stage discovery), the context lives in the log.Errorf line, and return err
  passes the raw error up.
  Do you prefer we add wrapping to system errors in apply.go as well? If so, should we apply the same change to
  export.go and transform.go for consistency, or keep apply different from the others?

Comment (msajidmansoori12) [reply cmd/apply/apply.go]: Thanks for checking — export does use `log.Errorf` then `return err` on those filesystem paths, that’s true.

The difference is apply already wrapped those three errors. Export has been a raw `return err` for a long time (even before the logging PRs). Transform still wraps DiscoverStages with `failed to discover stages: %w`, so this isn’t a single shared rule.

I tried the overwrite path locally with this PR. Cobra’s `Error:` line is only the syscall:

```text
level=error msg="Failed to clear output directory \"/tmp/.../blocked-out\": unlinkat ...: permission denied"
Error: unlinkat /tmp/.../blocked-out/keep: permission denied
```

On main that second line is:

```text
Error: failed to clear output directory: unlinkat ...: permission denied
```

The log line is useful, but `Error:` is what users (and scripts) see as the command failure. That’s the bit we lose if we drop the wrap.

Could you restore wrapping in apply only, same as before? No need to change export/transform in this PR.

cmd/apply/apply.go
Tamar-Dinavetsky: Hi @msajidmansoori12 , plan to use `o.log `in a follow-up when we add audit file configuration in `Complete`, so keeping it for now.

cmd/apply/apply.go
Tamar-Dinavetsky: Ok got it, thanks. 
Updated.

msajidmansoori12 (approved):
lgtm, thanks Tammy.

PR #789, opened 2026-08-11
:book: Document StatefulSet StorageClass conversion known limitation

### LINKED ISSUE #795: :book: Known limitation: StatefulSet scale-up after transfer-pvc StorageClass conversion uses original StorageClass
Description:

crane transfer-pvc supports migrating a PVC's data to a new PersistentVolumeClaim on a different StorageClass (commonly used for storage class conversion). However, this only affects the PVC object itself, it does not update the StorageClass referenced in the owning StatefulSet's volumeClaimTemplates.

Impact:

If a StatefulSet is scaled up after such a conversion (without being recreated), Kubernetes provisions the new replica's PVC directly from the StatefulSet's existing volumeClaimTemplates, which still references the original StorageClass. This happens silently, with no error or warning, resulting in inconsistent storage classes across replicas (e.g., data-mysql-0 on the new class, data-mysql-1 on the old one).

Root cause:

Kubernetes does not allow in-place updates to a StatefulSet's volumeClaimTemplates field, so there is no way for crane transfer-pvc (or any tool operating at the PVC level) to update it directly.

Workaround:

To fully complete a StorageClass conversion for a StatefulSet-managed application:

Delete the StatefulSet with --cascade=orphan (this removes only the StatefulSet controller object; existing pods and PVCs are left running/untouched).

Recreate the StatefulSet from a manifest with the updated storageClassName in volumeClaimTemplates.

Kubernetes re-adopts the existing pods/PVCs by name; future scale-ups will now correctly use the new StorageClass.

midays:
Documents a known limitation: `crane transfer-pvc` migrates PVC data to a new StorageClass but does not update the StatefulSet's `volumeClaimTemplates`. Scaling up after conversion without recreating the StatefulSet provisions new replicas on the original StorageClass.

## Summary by CodeRabbit

* **Documentation**
  * Added warnings that PVC migration does not update StatefulSet storage templates.
  * Documented the required procedure to recreate StatefulSets with the destination StorageClass before scaling.
  * Updated pre-migration checklists with this requirement.

### DIFF
--- a/docs/CRANE_COMPATIBILITY_MATRIX.md
+++ b/docs/CRANE_COMPATIBILITY_MATRIX.md
@@ -43,7 +43,7 @@ These resources are the core of any migration and are moved automatically.
   > **Warning — Namespace renaming:** If the namespace is renamed during migration, NetworkPolicy `namespaceSelector` entries that match the old namespace by label (e.g., `kubernetes.io/metadata.name: old-ns`) are not updated automatically. Manually update them before applying.
 * **Config & Secrets:** ConfigMap, Secret, ServiceAccount.
 * **Storage:** PersistentVolumeClaim (PVC).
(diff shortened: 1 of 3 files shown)


stillalearner (approved):
LGTM

PR #823, opened 2026-08-17
:sparkles: added encryption support for indirect migration

stillalearner:
## Summary

  Wire up the `--encrypt` flag for indirect PVC transfers. When set, crane auto-generates an encryption password, appends an rclone crypt overlay section to the rclone config, and creates the Secret with encryption enabled on both clusters.

  ### Changes

  - **`cmd/transfer-pvc/indirect.go`**
    - **`runIndirect()`** — When `--encrypt` is set, reads the rclone config file, generates a crypt section with an auto-generated password, appends it to the config data, then creates the Secret on both clusters with
  the combined config. Without `--encrypt`, behavior is unchanged.
    - **`createTempRcloneSecret` → `createTempRcloneSecretFromData`** — Renamed. Accepts `[]byte` config data instead of a file path. File reading moved to the caller so the config can be modified (crypt section
  appended) before creating the Secret.
    - **`generateCryptSection()`** — Generates a random 32-byte password, obscures it in rclone's format, and builds the `[encrypted]` crypt overlay config section pointing at the S3 path.
    - **`rcloneObscure()`** — Implements rclone's password obscuring format (AES-CTR with rclone's well-known key). Same algorithm as `rclone obscure` command, no rclone dependency needed.

  ### How it works

  When `--encrypt` is passed:

  1. Crane reads the user's rclone.conf
  2. Generates a random 32-byte encryption password
  3. Obscures it in rclone's format (AES-CTR + base64url)
  4. Appends an `[encrypted]` crypt overlay section to the config
  5. Creates the same Secret on both clusters (same password for encrypt + decrypt)
  6. crane-lib uses `encrypted:` as the remote path instead of the direct S3 path

  The resulting rclone.conf in the Secret:

  ```ini
  [remote]
  type = s3
  ...user's config...

  [encrypted]
  type = crypt
  remote = remote:bucket/ns/pvc
  password = <auto-generated-obscured>
```
  The password is random per transfer & ephemeral, not stored anywhere. After download and cleanup, the encrypted data on S3 is deleted.

  Tested locally

  - With --encrypt: file names and content encrypted on S3, data intact on target after decrypt (checksums match)
  - Without --encrypt: behavior identical to before, plaintext on S3

## Summary by CodeRabbit

* **New Features**
  * Added optional AES encryption for indirect transfers.
  * Automatically generates and securely obscures encryption passwords.
  * Includes encryption settings when creating transfer secrets.

* **Improvements**
  * Streamlined configuration handling for transfer setup.
  * Updated the underlying transfer library.
  * Added validation to require a local configuration file when encryption is enabled.

### DIFF
--- a/cmd/transfer-pvc/indirect.go
+++ b/cmd/transfer-pvc/indirect.go
@@ -370,7 +370,7 @@ func generateCryptSection(cloudStoragePath string) (string, error) {
 		return "", fmt.Errorf("failed to obscure encryption password: %w", err)
 	}
 
-	return indirect.BuildCryptSection(cloudStoragePath, obscured), nil
+	return indirect.BuildCryptSection(cloudStoragePath, obscured)
 }
 
 // rcloneObscure encodes a plaintext password in rclone's obscured format.
--- a/cmd/transfer-pvc/indirect.go
+++ b/cmd/transfer-pvc/indirect.go
@@ -2,7 +2,12 @@ package transfer_pvc
 
 import (
 	"context"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"strings"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: cmd/transfer-pvc/indirect.go [hunk 1]
```diff
@@ -71,13 +76,27 @@ func (t *TransferPVCCommand) runIndirect() error {
...
+		}
+
+		if t.Flags.Encrypt {
+			remotePath := fmt.Sprintf("%s/%s/%s", t.Flags.CloudStorage, t.PVC.Namespace.source, t.PVC.Name.source)
+			cryptSection, err := generateCryptSection(remotePath)
+			if err != nil {
+				return fmt.Errorf("failed to generate encryption config: %w", err)
+			}
+			configData = append(configData, []byte(cryptSection)...)
+		}
```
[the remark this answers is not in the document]
stillalearner: not needed.

cmd/transfer-pvc/indirect.go
```diff
@@ -344,3 +356,43 @@ func (t *TransferPVCCommand) createTempRcloneSecret(c client.Client, namespace,
...
+	obscured, err := rcloneObscure(base64.RawURLEncoding.EncodeToString(password))
+	if err != nil {
+		return "", fmt.Errorf("failed to obscure encryption password: %w", err)
+	}
+
+	return indirect.BuildCryptSection(cloudStoragePath, obscured)
+}
+
+// rcloneObscure encodes a plaintext password in rclone's obscured format.
+// This is AES-CTR with rclone's well-known key, identical to "rclone obscure".
```
aufi, line 381: 👍

cmd/transfer-pvc/indirect.go
```diff
@@ -71,13 +76,28 @@ func (t *TransferPVCCommand) runIndirect() error {
...
+		}
+
+		if t.Flags.Encrypt {
+			remotePath := fmt.Sprintf("%s/%s/%s", t.Flags.CloudStorage, t.PVC.Namespace.source, t.PVC.Name.source)
+			cryptSection, err := generateCryptSection(remotePath)
+			if err != nil {
+				return fmt.Errorf("failed to generate encryption config: %w", err)
+			}
+			configData = append(configData, '\n')
+			configData = append(configData, []byte(cryptSection)...)
```
aufi, line 94: I'm not sure if there is a reasonable chance to have already existing `[encrypted]` section, so it might become duplicated, but that's likely not a blocker, up to your decision.

aufi (commented):
Left few minor comments, but LGTM.

One more question is about (not)encrypting files metadata in the cloud storage, if those are not encrypted, we might need document it.

aufi (approved):
Left few minor comments, but LGTM.

One more question is about (not)encrypting files metadata in the cloud storage, if those are not encrypted, we might need document it.

cmd/transfer-pvc/indirect.go
stillalearner: ACK

stillalearner:
[review-docs]

stillalearner:
[update-docs]

PR #856, opened 2026-08-19
:bug: fix EndpointFlags.Validate pointer receiver to persist nginx default

### LINKED ISSUE #821: [BUG] Empty endpoint type defaults to nginx in Validate() but the default is lost
## Description
  `EndpointFlags.Validate()` has a value receiver. When the endpoint type is empty, it assigns the nginx
  default (`e.Type = endpointNginx`) — but since the receiver is a copy, that change is lost once `Validate()`
  returns. The caller is left with `Type == ""`.

## Impact
  An empty endpoint type with a valid subdomain passes validation, then later fails in `createEndpoint` because
  the type was never actually set to nginx.

  ## Steps to reproduce
  1. Set endpoint type to `""` and provide a valid subdomain.
  2. `Validate()` returns no error.
  3. `createEndpoint` receives an empty type and fails.

  ## Expected
  The nginx default should persist after `Validate()` so `Type == endpointNginx`.

## Additional info
[you can see the original message here](https://github.com/migtools/crane/pull/355#discussion_r3795368590)

Tamar-Dinavetsky:
Fixes https://github.com/migtools/crane/issues/821

## Summary
  
  `EndpointFlags.Validate()` used a value receiver, so the nginx default assigned inside (`e.Type =
  endpointNginx`) was lost after the method returned. The caller was left with `Type == ""`, causing

  ## Changes
  
  - Changed `Validate()` receiver from value to pointer (`*EndpointFlags`) so the nginx default persists after the
  call
  - Added a regression test (`TestEndpointFlags_Validate_DefaultPersists`) verifying that an empty `Type` with a
  valid subdomain correctly defaults to `nginx-ingress` after `Validate()`

## Summary by CodeRabbit

* **Bug Fixes**
  * Default endpoint settings are now correctly preserved when validating transfer options with a valid subdomain.
  * Prevents endpoint type defaults from being lost during validation.

### DIFF
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -97,7 +97,7 @@ type EndpointFlags struct {
 	IngressClass string
 }
 
-func (e EndpointFlags) Validate() error {
+func (e *EndpointFlags) Validate() error {
 	// default endpoint type is nginx-ingress
 	if e.Type == "" {
 		e.Type = endpointNginx
--- a/cmd/transfer-pvc/transfer-pvc_test.go
+++ b/cmd/transfer-pvc/transfer-pvc_test.go
@@ -18,6 +18,23 @@ import (
 
 func int64Ptr(v int64) *int64 { return &v }
 
+// TestEndpointFlags_Validate_DefaultPersists proves the value-receiver bug in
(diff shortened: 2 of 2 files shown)


aufi (approved):
The change looks good to me.

Issue #855, opened 2026-08-19
[BUG] [transfer-pvc] --keep-cloud-data silently ignored without --cloud-storage


nachandr:
**What version of crane are you running, and what are your clutsters+platform**
main

**What did you expect to happen?**
```Error: --keep-cloud-data requires --cloud-storage```

**What actually happened?**
```Error: subdomain cannot be empty when using nginx ingress```

**Please include any relevant logs or errors**

**Steps to reproduce:**
```bash
crane transfer-pvc \
  --source-context=kind-source-cluster \
  --destination-context=kind-dest-cluster \
  --pvc-name=test-pvc --pvc-namespace=test-indirect \
  --keep-cloud-data
```

PR #860, opened 2026-08-20
  :bug: Validate --keep-cloud-data requires --cloud-storage

Tamar-Dinavetsky:
https://github.com/migtools/crane/issues/855

## Summary

  causing a confusing unrelated error instead of a clear validation message.

  - Added early validation in `Validate()` that returns an explicit error when `--keep-cloud-data` is set without `--cloud-storage`
  - Placed before context evaluation so the error is shown even with non-existent kubeconfig contexts

 

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented invalid transfer configurations by rejecting `--keep-cloud-data` when cloud storage is not enabled.
  * Allowed `--keep-cloud-data` when valid cloud storage settings are provided.


stillalearner (approved):
LGTM

Issue #781, opened 2026-08-09
improve log messages in transfer-pvc directory.


Tamar-Dinavetsky:
parent: https://github.com/migtools/crane/issues/297

PR #862, opened 2026-08-20
:sparkles: feat(audit): improve logging in transfer-pvc command

Tamar-Dinavetsky:
https://github.com/migtools/crane/issues/781
Adds missing log statements across transfer-pvc commands as a first step toward the persistent audit log feature.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced progress and diagnostic logging throughout PVC transfers.
  * Added clearer debug, warning, and error messages for transfer phases, pod monitoring, cleanup, encryption, and partial outcomes.
  * Integrated transfer-PVC operations with the application’s shared logging configuration.
  * Transfer operations now consistently use the selected source image for setup and execution.
  * Improved reporting for configuration issues, security-context fallbacks, secret handling, and cleanup activities.

### DIFF
--- a/cmd/transfer-pvc/indirect.go
+++ b/cmd/transfer-pvc/indirect.go
@@ -8,8 +8,9 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
-	"log"
 	"os"
+
+	"github.com/sirupsen/logrus"
 	"strings"
 	"time"
 
@@ -25,6 +26,9 @@ import (
 )
 
 func (t *TransferPVCCommand) runIndirect() error {
+	log := t.globalFlags.GetLoggerOrDefault()
+	log.Infof("Starting indirect PVC transfer: %s/%s -> %s/%s", t.PVC.Namespace.source, t.PVC.Name.source, t.PVC.Namespace.destination, t.PVC.Name.destination)
+
 	fmt.Fprintf(os.Stderr, "\ncrane transfer-pvc (indirect via cloud storage)\n")
 	fmt.Fprintf(os.Stderr, "source context:      %s\n", t.Flags.SourceContext)
 	fmt.Fprintf(os.Stderr, "destination context: %s\n", t.Flags.DestinationContext)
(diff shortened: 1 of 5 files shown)


cmd/transfer-pvc/indirect.go
```diff
@@ -152,9 +164,10 @@ func (t *TransferPVCCommand) runIndirect() error {
 	fmt.Fprintf(os.Stderr, "[3/6] Uploading data to cloud storage ...\n")
 	uploadPod, err := transfer.Upload(context.TODO(), srcPVC)
 	if err != nil {
+		log.Debugf("Upload failed: %v", err)
```
msajidmansoori12, line 163: @Tamar-Dinavetsky 
Just wanted to point this out — I notice aborting failures on the indirect path are logged with Debugf (upload here, also client/PVC setup above and download a bit later), while the same kind of failure on the direct path uses Errorf.

Since the default log level is Info, these wouldn’t show up without --debug. The command still returns the error either way.

Can you please confirm if that was intentional? I think it should also be logged with errorf like direct path but maybe I missed something.

cmd/transfer-pvc/progress.go
```diff
@@ -500,7 +503,7 @@ func waitForPodRunning(c *kubernetes.Clientset, namespace string, labels map[str
 		}
 
 		if len(clientPodList.Items) != 1 {
-			log.Printf("expected 1 client pod found %d, with labels %v\n", len(clientPodList.Items), labels)
+			log.Debugf("Expected 1 client pod, found %d with labels %v", len(clientPodList.Items), labels)
```
msajidmansoori12: @Tamar-Dinavetsky  I noticed this wait loop used to print these with log.Printf (always visible), and they’re now Debugf.

This can sit here for a while waiting for the rsync client pod, so without --debug that wait would look silent. I think these should stay visible — maybe Warnf — since they’re the only signal that it’s still retrying, but just wanted to ask if the debug level was intentional to cut down the retry noise.

msajidmansoori12 (changes requested):

cmd/transfer-pvc/progress.go
Tamar-Dinavetsky: Thanks for catching that!

cmd/transfer-pvc/indirect.go
Tamar-Dinavetsky: Hi @msajidmansoori12 .The difference is intentional. the indirect path wraps errors with `fmt.Errorf `, while the direct path returns bare errors. So the log level follows the return type, not the severity.

CC: @stillalearner

msajidmansoori12: Thanks for the reply @Tamar-Dinavetsky .
I think I get what you mean — indirect already wraps with fmt.Errorf, so the user still sees the error even if the log is Debug.

I’m just not sure that should change the log level. Direct also adds a message with phases.Fail, and wrapping vs Debug/Error are separate things. Without --debug these still wouldn’t show in the audit log.

I think this should be Errorf like the direct path, but just wanted to check if Debug was still what you wanted.

@stillalearner  deferring it to you.

msajidmansoori12 (approved):
Approving so this isn’t blocked. The rsync wait logs look good.

Only leftover: aborting failures on the indirect path are still Debugf (direct uses Errorf), so they won’t show in the default audit log. Fine to follow up after merge if you still want Debug there.

cmd/transfer-pvc/indirect.go
msajidmansoori12: As confirmed by @stillalearner , fine to have it in Debug for now. Closing this thread.

Tamar-Dinavetsky: thanks @msajidmansoori12 for this review

PR #864, opened 2026-08-21
:sparkles: Filter resources GK in export

### LINKED ISSUE #863: Support for export resources filtering by GK
## Problem

Currently, crane export exports all namespaced resources (minus hard-coded Event skip) from a given namespace. There is no way to limit the export to specific resource types (e.g., only Deployments and ConfigMaps) or to exclude unwanted types.

For Buildconfig to Shipwright migration scenario it is expected to convert only relevant resources (primary BuildConfig) into the same cluster as Shipwright resources. In order to do that it is not handy have export full set of all existing resources in given namespace and trying import it back.

An option that Shipwright plugin would whiteout unrelated resources is not suitable since it would break possible combination of Shiwright plugin with Kubernetes/OpenShift plugins during migration (upgrade scenario) to another cluster while migrating to Shipwright.

##  Proposal

  Add support for filtering exported resources by Group/Kind (GK), using include and/or exclude semantics:
```
  crane export -n my-namespace --include-gk "BuildConfig" --include-gk "ConfigMap"
  crane export -n my-namespace --exclude-gk "Secret" --exclude-gk "Event"
```
##  Suggested flags

  - --include-gk (repeatable) — only export resources matching the specified Group/Kind values. When set, all non-matching types are skipped.
  - --exclude-gk (repeatable) — skip resources matching the specified Group/Kind values. All other types are exported.
  - Using both flags simultaneously should be a validation error.

aufi:
This PR adds --include-gk and --exclude-gk flags to filter exported resources by Group/Kind.

##  Supported formats

  - Kind — matches the Kind in any API group (e.g., Deployment matches apps/Deployment)
  - Group/Kind — matches the Kind in a specific API group (e.g., apps/Deployment, build.openshift.io/BuildConfig)

##  Backward compatibility

  Event resources are still filtered by default to maintain backward compatibility. The default behavior is equivalent to --exclude-gk Event. Users can override this by explicitly including Events with
  --include-gk Event.

 ## Implementation details

  - New GKFilter type in cmd/export/gk_filter.go handles parsing and filtering logic
  - --include-gk and --exclude-gk flags added to export command (mutually exclusive)
  - Default excludeGK = ["Event"] is set in Complete() when no GK filters are specified, default explude could be changed with providing different excludeGK  flag content
  - Filter is applied in resourceToExtract() before API calls are made
  - Works alongside existing --label-selector filtering

##  Examples
```
  Include only specific resource types

  # Export only Deployments and ConfigMaps
  crane export -n my-app --include-gk Deployment --include-gk ConfigMap

  # Export only OpenShift BuildConfigs
  crane export -n my-app --include-gk build.openshift.io/BuildConfig

  # Export Deployments from any group (apps, extensions, etc.)
  crane export -n my-app --include-gk Deployment

  Exclude specific resource types

  # Export everything except Secrets
  crane export -n my-app --exclude-gk Secret

  # Exclude multiple types
  crane export -n my-app --exclude-gk Secret --exclude-gk ConfigMap

  # Default behavior: Events are automatically excluded
  crane export -n my-app
  # Equivalent to: crane export -n my-app --exclude-gk Event

  Override default Event exclusion

  # Explicitly include Events
  crane export -n my-app --include-gk Event

  Group-specific filtering

  # Only export apps/Deployment (not extensions/Deployment)
  crane export -n my-app --include-gk apps/Deployment

  # Exclude core API Pods specifically (using empty group)
  crane export -n my-app --exclude-gk /Pod
```

Docs will follow up in separate PR.

Fixes https://github.com/migtools/crane/issues/863

## Summary by CodeRabbit

* **New Features**
  * Added Group/Kind filtering for exported resources.
  * Supports include-only and exclude filters by kind, group/kind, or core group.
  * Events remain excluded by default unless explicitly included.
* **Bug Fixes**
  * Improved validation for conflicting or incorrectly formatted filters.
  * Enabled explicit inclusion of Event resources when requested.
  * Preserved access to cluster-scoped resources during filtering.


cmd/export/export.go
```diff
@@ -101,6 +103,13 @@ func (o *ExportOptions) Complete(c *cobra.Command, args []string) error {
 		}
 	}
 
+	// Apply default --exclude-gk Event if no GK filters specified
+	// This maintains backward compatibility: Events are skipped by default
+	// Users can override by explicitly using --include-gk Event or --exclude-gk <other-kinds>
+	if len(o.includeGK) == 0 && len(o.excludeGK) == 0 {
```
msajidmansoori12, line 109: Hi @aufi,
 I tested this locally and wanted to double-check the intended behavior here.
With no GK flags, Events are excluded as expected, but when I run --exclude-gk Secret, Events get exported again unless Event is also explicitly excluded. 
My understanding from the PR description was that Events should stay excluded by default for backward compatibility, so I expected --exclude-gk Secret to behave more like adding to the default exclusion rather than replacing it.
Could you please confirm whether this is intentional?

msajidmansoori12 (changes requested):

cmd/export/export.go
aufi: Thanks for testing it, this was intentional, wanted to make possible remove Events from excluded list when something was specified. Will update decription/docs in followup docs PR.

msajidmansoori12: Thanks for clarifying @aufi !

msajidmansoori12 (approved):
lgtm , thanks Marek !

PR #875, opened 2026-08-25
:test_tube: automate mta901

stillalearner:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of same-cluster storage migrations involving PVC renaming and StorageClass conversion.
  * Enhanced validation of renamed PVC bindings and migration mappings.
  * Added checks to confirm temporary migration resources are removed after completion.
* **Tests**
  * Added end-to-end coverage for MongoDB migrations with PVC renaming and StorageClass conversion.
  * Expanded validation of migrated data, resource cleanup, and destination storage binding.

### DIFF
--- a/e2e-tests/framework/pipeline.go
+++ b/e2e-tests/framework/pipeline.go
@@ -168,17 +168,22 @@ func ApplyOutputToTargetWithNamespaceRemapNonAdmin(kubectl KubectlRunner, srcNam
 	return nil
 }
 
-// VerifyPVCReferenceInOutput checks that the rendered output.yaml contains
-// the expected PVC name. Useful for verifying pvc-rename-map transforms
-// updated the workload's claimName after crane apply.
-func VerifyPVCReferenceInOutput(outputDir, expectedPVCName string) {
+// VerifyPVCRenameInOutput checks that the rendered output.yaml contains
+// the new PVC name and does not contain the old PVC name as a claimName.
+// This verifies the pvc-rename-map transform both added the new reference
+// and removed the old one.
+func VerifyPVCRenameInOutput(outputDir, oldPVCName, newPVCName string) {
 	outputFile := filepath.Join(outputDir, "output.yaml")
 	content, err := os.ReadFile(outputFile)
 	gomega.Expect(err).NotTo(gomega.HaveOccurred(), "failed to read %s", outputFile)
-	log.Printf("Checking output.yaml for PVC reference %q", expectedPVCName)
+	text := string(content)
 
-	gomega.Expect(string(content)).To(gomega.ContainSubstring(expectedPVCName),
-		"output.yaml should reference the renamed PVC %q", expectedPVCName)
+	log.Printf("Checking output.yaml: expect claimName %q, reject claimName %q", newPVCName, oldPVCName)
+
+	gomega.Expect(text).To(gomega.ContainSubstring("claimName: "+newPVCName),
+		"output.yaml should contain claimName: %s", newPVCName)
+	gomega.Expect(text).NotTo(gomega.ContainSubstring("claimName: "+oldPVCName),
+		"output.yaml should not contain old claimName: %s (transform should have renamed it)", oldPVCName)
 }
 
 // AssertNoTransferPVCLeftovers waits until transfer-pvc helper objects labeled
--- a/e2e-tests/framework/pipeline.go
+++ b/e2e-tests/framework/pipeline.go
@@ -168,6 +168,19 @@ func ApplyOutputToTargetWithNamespaceRemapNonAdmin(kubectl KubectlRunner, srcNam
 	return nil
 }
 
+// VerifyPVCReferenceInOutput checks that the rendered output.yaml contains
+// the expected PVC name. Useful for verifying pvc-rename-map transforms
+// updated the workload's claimName after crane apply.
+func VerifyPVCReferenceInOutput(outputDir, expectedPVCName string) {
+	outputFile := filepath.Join(outputDir, "output.yaml")
+	content, err := os.ReadFile(outputFile)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred(), "failed to read %s", outputFile)
+	log.Printf("Checking output.yaml for PVC reference %q", expectedPVCName)
+
+	gomega.Expect(string(content)).To(gomega.ContainSubstring(expectedPVCName),
+		"output.yaml should reference the renamed PVC %q", expectedPVCName)
+}
+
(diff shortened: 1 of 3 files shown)


e2e-tests/framework/pipeline.go
```diff
@@ -168,6 +168,19 @@ func ApplyOutputToTargetWithNamespaceRemapNonAdmin(kubectl KubectlRunner, srcNam
...
+// the expected PVC name. Useful for verifying pvc-rename-map transforms
+// updated the workload's claimName after crane apply.
+func VerifyPVCReferenceInOutput(outputDir, expectedPVCName string) {
+	outputFile := filepath.Join(outputDir, "output.yaml")
+	content, err := os.ReadFile(outputFile)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred(), "failed to read %s", outputFile)
+	log.Printf("Checking output.yaml for PVC reference %q", expectedPVCName)
+
+	gomega.Expect(string(content)).To(gomega.ContainSubstring(expectedPVCName),
+		"output.yaml should reference the renamed PVC %q", expectedPVCName)
```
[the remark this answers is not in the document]
stillalearner: The PVC name (mongodb-data-new) is unique enough that a false positive from an unrelated field is practically impossible

e2e-tests/tests/tier0/mta_901_same_cluster_pvc_rename_storageclass_conversion_test.go
```diff
@@ -0,0 +1,156 @@
...
+		Expect(kubectlTgt.ScaleDeployment(tgtNamespace, appName, 1)).NotTo(HaveOccurred())
+		Eventually(tgtApp.Validate, "5m", "10s").Should(Succeed())
+
+		By("Assert destination PVC is Bound after the workload starts")
+		destPVC, err = GetPVC(tgtApp.Context, tgtNamespace, dstPVCName)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(destPVC.Status.Phase).To(Equal(corev1.ClaimBound))
+
+		By("Confirm no leftover transfer-pvc resources")
+		AssertNoTransferPVCLeftovers(kubectlTgt, []string{srcNamespace, tgtNamespace}, dstPVCName)
```
msajidmansoori12: Hi @stillalearner , I think this leftover check is using the wrong PVC name. In this flow, the temporary transfer-pvc resources are labeled with the source PVC name, but the test is checking with dstPVCName. Because of that, the check may pass even if some temporary resources are still left behind. I think this should use srcPVCName so the test is actually verifying cleanup.

e2e-tests/tests/tier0/mta_901_same_cluster_pvc_rename_storageclass_conversion_test.go
```diff
@@ -0,0 +1,156 @@
...
+		Expect(PVCStorageClassName(*destPVC)).To(Equal(destSCName),
+			"destination PVC StorageClass should be %s", destSCName)
+		log.Printf("Destination PVC %s/%s created with StorageClass=%s", tgtNamespace, dstPVCName, destSCName)
+
+		By("Run crane export/transform/apply pipeline with pvc-rename-map")
+		runner.WorkDir = paths.TempDir
+		Expect(RunCranePipelineWithChecks(runner, exportOpts, transformOpts, applyOpts)).NotTo(HaveOccurred())
+
+		By("Verify the rendered output references the new PVC name")
+		VerifyPVCReferenceInOutput(paths.OutputDir, dstPVCName)
```
msajidmansoori12: @stillalearner  I think this is a bit weaker than the test steps intend. The steps say we should inspect the transformed manifest directly before apply and verify the PVC reference changed from the old name to the new one. Here we run the full pipeline first and then only check that output.yaml contains the new PVC name, which may miss cases where transform did not rewrite the workload reference correctly. 

Deferring it to you , if you think this is enough , please update the test steps accordingly.

msajidmansoori12 (changes requested):

e2e-tests/tests/tier0/mta_901_same_cluster_pvc_rename_storageclass_conversion_test.go
stillalearner: ack

e2e-tests/tests/tier0/mta_901_same_cluster_pvc_rename_storageclass_conversion_test.go
stillalearner: ya i understand, though if mapping worked, i think its fine to check this much.

stillalearner: still, let me make it tighter

e2e-tests/tests/tier0/mta_901_same_cluster_pvc_rename_storageclass_conversion_test.go
```diff
@@ -0,0 +1,158 @@
...
+		Eventually(tgtApp.Validate, "5m", "10s").Should(Succeed())
+
+		By("Assert destination PVC is Bound after the workload starts")
+		destPVC, err = GetPVC(tgtApp.Context, tgtNamespace, dstPVCName)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(destPVC.Status.Phase).To(Equal(corev1.ClaimBound))
+
+		By("Confirm no leftover transfer-pvc resources")
+		AssertNoTransferPVCLeftovers(kubectlSrc, []string{srcNamespace}, srcPVCName)
+		AssertNoTransferPVCLeftovers(kubectlTgt, []string{tgtNamespace}, dstPVCName)
```
msajidmansoori12: Hi Sachin, I think this target-side leftover check is still using the wrong PVC name. In the same-cluster different-namespace flow, transfer-pvc starts with the source PVC name in the helper-resource label, and only changes it to the destination PVC name in the same-namespace special case:


```
labels := map[string]string{
    "app.konveyor.io/created-for-pvc": getValidatedResourceName(srcPVC.Name),
}
if t.isIntraClusterSameNamespace() {
    labels["app.konveyor.io/created-for-pvc"] = getValidatedResourceName(destPVC.Name)
}
```

Because of that, checking the target namespace with dstPVCName can still miss leftover temporary resources. I think this should still use srcPVCName here.

msajidmansoori12 (changes requested):

msajidmansoori12 (approved):
lgtm , thanks !

PR #882, opened 2026-08-26
:test_tube: transfer-pvc : Fail fast on a missing rclone config Secret during indirect migration

### LINKED ISSUE #892: tier1: Fail fast on a missing rclone config Secret

nachandr:
Fixes [892](https://github.com/migtools/crane/issues/892)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage confirming that PVC transfers fail clearly when the referenced rclone configuration Secret is missing.
  * Added validation that the error identifies the missing Secret.
  * Added checks ensuring failed transfers do not leave behind orphaned resources.


e2e-tests/tests/tier1/indirect_migration/e2e_suite_test.go
```diff
@@ -0,0 +1,41 @@
+package indirect_migration
```
msajidmansoori12: Hi @nachandr ,
Trying to understand if the new tier1/indirect_migration/ package is a strict requirement for this test, or if I’m missing something — could this live in tier1/ like the other mta_* tests?

msajidmansoori12 (changes requested):

e2e-tests/tests/tier1/mta_910_indirect_missing_rclone_secret_test.go
```diff
@@ -0,0 +1,121 @@
+package indirect_migration
```
msajidmansoori12: Hi Nandini — one more thing on intent: direct CI also runs --label-filter="tier1" without --cloud-storage, but this test hardcodes CloudStorage in the opts, so it still runs indirect transfer-pvc on direct runs too. Was that intentional, or should we skip when config.CloudStorage == "" (opposite of MTA-877) so it only runs in the indirect workflow?

msajidmansoori12 (changes requested):

e2e-tests/tests/tier1/mta_910_indirect_missing_rclone_secret_test.go
nachandr: This test should only run for the indirect mode.

e2e-tests/tests/tier1/indirect_migration/e2e_suite_test.go
nachandr: There are several tests in the tier1 directory, so I'm organizing the tests based on the feature. It could still live in the tier1/ directory .

msajidmansoori12: @nachandr Indirect CI currently runs the full tier1 suite (--label-filter="tier1"), not just this package — so the new folder doesn’t really define an “indirect migration” run today. If we wanted the folder to mean that, we’d either need to move/copy the other indirect-related tier1 tests here (which feels redundant) or change CI to target this package/label separately.

I would prefer to keep this in tier1 directory for now , because for indirect runs we are running all the tier1 tests.

e2e-tests/tests/tier1/mta_910_indirect_missing_rclone_secret_test.go
msajidmansoori12: For that you can do something like this - 
https://github.com/migtools/crane/blob/main/e2e-tests/tests/tier1/mta_877_dvm_limitrange_test.go#L26 

Just check the opposite , something like` if config.CloudStorage == "" { skip("") }` so that for direct runs which don't provide this flag skips this test.

msajidmansoori12 (changes requested):

msajidmansoori12 (approved):
lgtm , thanks @nachandr

PR #897, opened 2026-08-31
:test_tube: Automate indirect data transfer with --rclone-config-secret

### LINKED ISSUE #893: tier1 : Indirect data transfer with --rlcone-config-secret

nachandr:
Fixes [893](https://github.com/migtools/crane/issues/893)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for indirect persistent volume migration through S3 using an existing rclone configuration Secret.
  * Verifies that migrated data remains intact after transfer.
  * Confirms the provided Secret’s identity, contents, and labels remain unchanged on both clusters.
  * Covers transfers performed by a non-administrator user.
  * Added suite configuration and command-line options for running indirect migration tests across supported environments.


stillalearner (changes requested):
I dont think a separate suite is needed here. 

We can move this to existing tier1 suite only.

Is there a specific reason? 

If its only classification, we can find another way using labels, if it is an indirect_only test case.

stillalearner (approved):

PR #919, opened 2026-09-03
:bug: Fix dest pvc storage class check if existing

### LINKED ISSUE #894: [BUG] transfer-pvc uses pre-existing destination PVC with wrong StorageClass
### Description

  When `--dest-storage-class` is specified but the destination PVC already exists with a different StorageClass, crane skips PVC creation (treats `AlreadyExists` as success) and proceeds to copy data into the wrong-class
  PVC. No error, no warning, exit 0.

  The user believes the StorageClass conversion succeeded, but the data landed on the original StorageClass.

  ### Steps to reproduce

  1. Create source PVC on `standard` StorageClass with data
  2. Pre-create destination PVC on `crane-sc02-target` StorageClass (wrong class)
  3. Run transfer requesting `standard-v2`:

  ```bash
  crane transfer-pvc \
    --source-context src --destination-context tgt \
    --pvc-name test-pvc --pvc-namespace myns \
    --dest-storage-class standard-v2 \
    --endpoint route
```
 ## Expected

  Crane should detect the StorageClass mismatch and fail with a clear error:

  Error: destination PVC "test-pvc" already exists with StorageClass "crane-sc02-target"
  but --dest-storage-class "standard-v2" was requested. Delete the existing PVC or
  omit --dest-storage-class to use the existing PVC as-is.

##  Actual

  - [2/7] Creating destination PVC ... ok (silently skipped)
  - Data copied to the wrong-class PVC
  - PVC data copy: succeeded
  - Exit code: 0

  The destination PVC remains on crane-sc02-target — the requested standard-v2 conversion never happened.

 ## Root cause

  // cmd/transfer-pvc/transfer-pvc.go
  err = destClient.Create(context.TODO(), destPVC, &client.CreateOptions{})
  if err != nil && !errors.IsAlreadyExists(err) {
      return fmt.Errorf("unable to create destination PVC: %w", err)
  }

  AlreadyExists is treated as unconditional success. No validation that the existing PVC's StorageClass, access mode, or capacity matches the requested parameters.

 ## Impact

  - StorageClass conversion silently fails — user thinks data is on the new class but it's on the old one

##  Environment

  - Crane commit: 8ca0870
  - Kubernetes v1.34.0 (minikube)
  - Reproducible on both minikube and OCP

### PR DESCRIPTION
**Pull request by stillalearner**
Fixes #894 

## Bug Fixes

* Improved destination PVC handling during transfers.
* Existing destination PVCs are accepted when their storage class matches the requested value—or when no storage class is specified.
* Transfers now report a clear error when an existing PVC’s storage class differs from the requested value, with guidance to resolve the conflict.

### DIFF
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -1290,6 +1290,8 @@ func (t *TransferPVCCommand) buildDestinationPVC(sourcePVC *corev1.PersistentVol
 	return pvc
 }
 
+// createDestinationPVC creates the destination PVC, validating the storage
+// class of an existing PVC when --dest-storage-class was requested.
 func (t *TransferPVCCommand) createDestinationPVC(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) error {
 	if err := c.Create(ctx, pvc, &client.CreateOptions{}); err == nil {
 		return nil
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -421,8 +421,8 @@ func (t *TransferPVCCommand) run() (retErr error) {
 	log.Infof("Phase 2: Creating destination PVC")
 	phases.Start("Creating destination PVC")
 	destPVC := t.buildDestinationPVC(srcPVC)
-	err = destClient.Create(context.TODO(), destPVC, &client.CreateOptions{})
-	if err != nil && !errors.IsAlreadyExists(err) {
+	err = t.createDestinationPVC(context.TODO(), destClient, destPVC)
+	if err != nil {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cmd/transfer-pvc/transfer-pvc.go [hunk 1]
```diff
@@ -1290,6 +1290,33 @@ func (t *TransferPVCCommand) buildDestinationPVC(sourcePVC *corev1.PersistentVol
 	return pvc
 }
 
+func (t *TransferPVCCommand) createDestinationPVC(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) error {
```
msajidmansoori12, line 1295: @stillalearner perhaps good to have docstring comment for this function. Rest looks good to me !

msajidmansoori12 (approved):

aufi (approved):

PR #940, opened 2026-09-08
:bug: Fix per-stage optionals in instructions file

### LINKED ISSUE #925: Transform instructions-file fails with per-stage optionals
## Summary

`crane transform --instructions-file` cannot complete an instructions file
that contains **more than one stage** and **at least one `optionals`** map. The command
runs instruction stages one at a time, but validates the complete map of
per-stage optionals against only the stage currently being run.

## Environment

- `crane version`:

  ```text
  crane:
      Version: main
      SHA: main
  crane-lib:
      Version: v0.0.10
  kustomize:
      Version: v5.8.1
  ```

- Built-in plugins reported by `crane transform list-plugins`:
  `KubernetesPlugin`, `OpenShiftPlugin`, `BuildConfigPlugin`.
- Tested against the WordPress workload in namespace
  `wordpress-instructions-demo` on OpenShift Local (CRC).

## Reproducer

Run crane transform with instructions file:

```yaml
stages:
  - name: KubernetesPlugin
    optionals:
      add-annotations: "migrated-by=crane,migration-demo=multi-stage-optionals"
      strip-default-rbac: "true"
  - name: OpenShiftPlugin
    optionals:
      strip-default-rbac: "true"
      strip-default-cabundle: "true"
```

## Actual Result

The first stage fails because `OpenShiftPlugin` is not part of the selector for
the currently running `KubernetesPlugin` stage:

```text
Invalid stage optional flags: per-stage optionals reference unknown stage "OpenShiftPlugin" (known stages: KubernetesPlugin)
```

Keeping optionals only on `KubernetesPlugin` does not work around the problem.
`KubernetesPlugin` completes, but the next `CustomEdits` stage fails with:

```text
Invalid stage optional flags: per-stage optionals reference unknown stage "KubernetesPlugin" (known stages: CustomEdits)
```

## Expected Result

Each stage should receive only its own `optionals` map. A multi-stage
instructions file with optionals for `KubernetesPlugin`, `OpenShiftPlugin`, and
`BuildConfigPlugin` should run all stages in declared order.

## Impact

The instructions-file schema advertises per-stage optionals, but they cannot be
used in any multi-stage pipeline. The runnable `instructions.yaml` sample is
therefore intentionally option-free.

## Suggested Fix

When executing each instruction stage, pass only the optional values keyed by
that stage to the orchestrator, or validate `StageOptionalFlags` against the
complete instructions stage list rather than the one-stage selector.

aufi:
Limit transform instruction-file optionals to the stage currently being executed, preventing validation failures for optionals configured on other stages.

Add regression coverage for stage-specific optional flag isolation.

Fixes https://github.com/migtools/crane/issues/925

## Summary by CodeRabbit

- **Bug Fixes**
  - Instruction stages now run with only the optional settings relevant to their selected plugin.
  - Stage names and generated stage directories are resolved independently for more reliable stage execution.

- **Tests**
  - Added coverage to verify plugin-specific optional settings are correctly selected, including cases where none are available.

### DIFF
--- a/cmd/transform/transform.go
+++ b/cmd/transform/transform.go
@@ -214,6 +214,7 @@ func (o *Options) run() error {
 	}
 
 	var instructionStages []string
+	var instructionPluginStages []string
 	var instructionStageOptionals map[string]map[string]string
 	if o.InstructionsFile != "" {
 		instructionsFilePath, err := filepath.Abs(o.InstructionsFile)
@@ -226,7 +227,8 @@ func (o *Options) run() error {
 			log.Errorf("Failed to load instructions file %q: %v", instructionsFilePath, err)
 			return err
 		}
-		instructionStages = internalTransform.GenerateStageDirNames(cfg.StageNames())
+		instructionPluginStages = cfg.StageNames()
+		instructionStages = internalTransform.GenerateStageDirNames(instructionPluginStages)
 		instructionStageOptionals, err = cfg.StageOptionals()
 		if err != nil {
 			return fmt.Errorf("invalid instructions file %q: %w", instructionsFilePath, err)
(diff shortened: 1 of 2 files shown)


msajidmansoori12 (approved):
lgtm @aufi , thanks !

PR #957, opened 2026-09-10
:book: Update transfer-pvc doc with info on file ownership

### LINKED ISSUE #969: [BUG] Update transfer-pvc doc with info on file ownership
The [Indirect migration enhancement](https://github.com/konveyor/enhancements/pull/301) doc states this:
Filesystem metadata (mode bits, ownership, xattrs) is preserved using rclone's --metadata flag.

However, in reality file ownership on target is overwritten to mover-pod UID/GID.

nachandr:
The [Indirect migration enhancement](https://github.com/konveyor/enhancements/pull/301) states this:
Filesystem metadata (mode bits, ownership, xattrs) is preserved using rclone's --metadata flag.

However, in reality  file ownership on target is overwritten to mover-pod UID/GID.

## Documentation

- Clarified that both direct and indirect PVC transfer modes preserve file contents, permissions, and directory structure.
- Documented that ownership is generally normalized when transfers use non-root mover Pods.
- Added details on ownership handling for direct transfers and UID/GID resolution in indirect transfers.
- Updated guidance for workloads requiring specific ownership or group access.

### DIFF
--- a/docs/commands/transfer-pvc.md
+++ b/docs/commands/transfer-pvc.md
@@ -180,7 +180,7 @@ Do not use `--rclone-config-file` with `--rclone-config-secret`.
 Preserving original UID/GID requires the receiving process to run as root, which crane's mover Pods normally do not. As a result, restored files generally do **not** keep their original owner/group. This affects both transfer modes, and changing UIDs is expected when moving data between clusters. The two modes differ in *how* ownership is resolved:
 
 - **Direct mode** runs rsync with `--owner`/`--group`. As root these preserve the original UID/GID; as a non-root Pod (the common case) rsync cannot restore arbitrary owners and silently maps files to the UID it runs as. The transfer still succeeds.
-- **Indirect mode** explicitly rewrites ownership. During download, crane-lib appends `--metadata-set uid=<N> --metadata-set gid=<N>` to the rclone `sync` command in the destination (mover) Pod, so files are always set to the resolved download security context, never the original owner. The values are derived as follows:
+- **Indirect mode** explicitly rewrites ownership. Internally (not a user-facing option), crane sets ownership metadata on every downloaded file so files always match the resolved download security context, not the original owner. The values are derived as follows:
   - **UID** is `RunAsUser` when it is set, otherwise it falls back to **65534 (nobody)**.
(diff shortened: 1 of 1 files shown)


aufi (commented):
I'm positive on documenting possible changes in files ownership, however I think reality is little different, than described and it looks little more as an information than limitation, but good to be mentioned. 

The difference as far I can understand it is primary in what is handled by direct rsync as implicit fallback of ownership to current user and indirect rclone, that sets requested metadata or fails without fallback (so, it is executed in way it should always succeed).
So, it affects both direct and indirect, only difference here is that direct rsync might keep ownerships (but more often/close to always it will not keep them when running as non-root IMO), indirect will keep them never.

Please keep me honest if the comment doesn't match to the crane behavior.

docs/commands/transfer-pvc.md
```diff
@@ -173,6 +173,25 @@ Do not use `--rclone-config-file` with `--rclone-config-secret`.
...
+
+##### Note: file ownership (UID/GID) is normalized on transfer
+
+Preserving original UID/GID requires the receiving process to run as root, which crane's mover Pods normally do not. As a result, restored files generally do **not** keep their original owner/group. This affects both transfer modes, and changing UIDs is expected when moving data between clusters. The two modes differ in *how* ownership is resolved:
+
+- **Direct mode** runs rsync with `--owner`/`--group`. As root these preserve the original UID/GID; as a non-root Pod (the common case) rsync cannot restore arbitrary owners and silently maps files to the UID it runs as. The transfer still succeeds.
+- **Indirect mode** explicitly rewrites ownership. During download, crane-lib appends `--metadata-set uid=<N> --metadata-set gid=<N>` to the rclone `sync` command in the destination (mover) Pod, so files are always set to the resolved download security context, never the original owner. The values are derived as follows:
```
aufi, line 183: The rest of this document contains commands used by end user, not executed internally by our tooling. Mentioning them explicitly might provide deep information to user (when interested), on the other hand might look misleading providing options, that user cannot directly use.

aufi (approved):
Overall LGTM. Left one comment, but likely not a blocker, handle it as you prefer.

docs/commands/transfer-pvc.md
nachandr: Hi @aufi , Good point — reworded to make clear ownership metadata is set internally and isn't a user-facing option, and dropped the raw rclone flags. Thanks!

PR #959, opened 2026-09-10
:test_tube: add test for pvc annotations after migration

stillalearner:
## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage confirming that user-defined persistent volume claim annotations remain preserved when an application is migrated.
  * Verified that migrated claims successfully bind using the destination storage configuration.
  * Confirmed that destination storage provisioning and binding annotations are applied correctly without overwriting user-defined annotations.
  * Improved validation of storage metadata consistency across the migration process.

### DIFF
--- a/e2e-tests/tests/tier1/mta_888_pvc_annotations_test.go
+++ b/e2e-tests/tests/tier1/mta_888_pvc_annotations_test.go
@@ -1,0 +1,132 @@
+package e2e
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/konveyor/crane/e2e-tests/config"
+	. "github.com/konveyor/crane/e2e-tests/framework"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	mta888AnnotationKey   = "e2e.crane.konveyor.io/retention"
+	mta888AnnotationValue = "keep-unchanged"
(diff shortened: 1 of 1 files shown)


e2e-tests/tests/tier1/mta_888_pvc_annotations_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/konveyor/crane/e2e-tests/config"
+	. "github.com/konveyor/crane/e2e-tests/framework"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
```
midays: These constants are defined at package level but are only used in the MTA-888 test:

```go
const (
    mta888AnnotationKey   = "e2e.crane.konveyor.io/retention"
    mta888AnnotationValue = "keep-unchanged"
    mta888PVCName         = "mongodb-data"
)
```

instead these should be moved to inside the test function

```go
It("[MTA-888] preserves user PVC annotations...", func() {
    const (
        annotationKey   = "e2e.crane.konveyor.io/retention"
        annotationValue = "keep-unchanged"
        pvcName         = "mongodb-data"
        appName         = "mongodb"
        namespace       = "mta-888-pvc-annotations"
    )
    // ...
})
```

midays (commented):
minor comment

midays (approved):

Issue #945, opened 2026-09-08
audit-log: fail with error when audit log file cannot be created


Tamar-Dinavetsky:
### Description:

Currently, if the `--audit-log` path cannot be created (e.g. bad permissions, non-existent parent directory), crane prints a warning and continues executing — but no audit log is written anywhere:

`level=warning msg="Failed to open audit log file /root/bad/path.log: mkdir /root: read-only file system"
`
This silently drops all audit data, which is unexpected.

### Proposed behavior:
Return an error from `Complete() `when the audit log file cannot be created, so the command fails before executing any operation. The warning is already logged before `Validate()` and `Run()`, so the check happens at the right point.

### Implementation note:
`GetLogger()` currently returns only `*logrus.Logger`. It needs to also return an error so `Complete()` can propagate it.

aufi:
Thanks for finding this issue, I also think it is a bug and agree with proposed solution.

Adding @stillalearner and @istein1 for visibility and to confirm described behaviour.

stillalearner:
@aufi do you think we should block execution of command, only because of audit log?

istein1:
@rromannissen ,

We recently added audit logging to Crane (PR #913) to give visibility into migration operations. 
A follow-up PR (#960) is now proposing that if the audit log file can't be created, 
Crane should block the command entirely rather than warn and continue silently.

The question is: should Crane refuse to run if it can't write audit logs?

The case for blocking is that if a user has opted into audit logging, running without it creates a silent gap in the audit trail — which could be a compliance concern. 
The case against is that it makes the tool less forgiving, especially for commands that don't actually change anything.

A middle ground that's common in tools like Linux auditd and Kubernetes audit policy:
- Fail hard on commands that mutate state (apply, transform, transfer-pvc, export) — these are the operations you'd actually need an audit trail for
- Warn and continue on read-only commands (plugin list, validate, shell completions) — no state changes, so the compliance risk is low
This way we protect the audit trail where it matters without making the tool unnecessarily strict for non-destructive operations.

What are your thoughts — should we go all-or-nothing, or differentiate by command type?

rromannissen:
@istein1 I'm ok with it failing if it doesn't have permission to write on the target location. I think logging is a core part of the user experience, and we shouldn't allow degradation where logs are not written. If the user is passing the target path via the --audit-log flag, that means there is an explicit interest on having the logs located somewhere, and I don't think not writing them or having a degraded version of them would be acceptable in that case.

PR #960, opened 2026-09-10
:bug: fix(audit): Return error when audit log file cannot be created

Tamar-Dinavetsky:
https://github.com/migtools/crane/issues/945

## Summary
- Make `GetLogger()` and `GetLoggerOrDefault()` return `(*logrus.Logger, error)` to propagate audit log creation failures
- Add error checks in all command `Complete()` methods to fail early if audit log cannot be initialized
- Handle completion functions specially (return `ShellCompDirectiveError` instead of propagating error)
- Update `initConfig()` to handle logger initialization error

## Fixes
Previously, when the audit log path couldn't be created, crane would warn but continue executing without writing any audit data. Now it fails before running the command.

## Summary by CodeRabbit

- **Bug Fixes**
  - Logger initialization errors are now preserved and reported consistently across commands.
  - Command execution stops when required logger setup fails, preventing operations with incomplete logging.
  - Configuration loading continues when audit logging cannot be initialized, while the issue is surfaced as a warning.
  - Audit logging is skipped when no audit log path is configured.
  - Merged configuration values are now applied correctly during command startup.
  - Logger setup now occurs consistently before command execution begins.

### DIFF
--- a/internal/flags/global_flags.go
+++ b/internal/flags/global_flags.go
@@ -36,9 +36,9 @@ func (g *GlobalFlags) SetCmdName(name string) {
 }
 
 // GetLoggerOrDefault returns the configured logger, or logrus.StandardLogger() if GlobalFlags is nil.
-func (g *GlobalFlags) GetLoggerOrDefault() *logrus.Logger {
+func (g *GlobalFlags) GetLoggerOrDefault() (*logrus.Logger, error) {
 	if g == nil {
-		return logrus.StandardLogger()
+		return logrus.StandardLogger(), nil
 	}
 	return g.GetLogger()
 }
@@ -49,7 +49,7 @@ func isCompletionMode() bool {
 	return len(os.Args) > 1 && (os.Args[1] == "__complete" || os.Args[1] == "__completeNoDesc")
 }
 
-func (g *GlobalFlags) GetLogger() *logrus.Logger {
+func (g *GlobalFlags) GetLogger() (*logrus.Logger, error) {
 	if g.logger == nil {
 		g.logger = logrus.New()
 		g.logger.SetLevel(logrus.DebugLevel)
@@ -63,10 +63,11 @@ func (g *GlobalFlags) GetLogger() *logrus.Logger {
 				g.logger.AddHook(fileHook)
 			} else {
 				g.logger.Warnf("Failed to open audit log file %s: %v", g.AuditLogPath, err)
+				return g.logger, err
 			}
 		}
 	}
-	return g.logger
+	return g.logger, nil
 }
 
 // Close releases the audit log file. Call this when the program exits.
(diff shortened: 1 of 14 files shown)


internal/flags/global_flags.go
```diff
@@ -49,7 +49,7 @@ func isCompletionMode() bool {
 	return len(os.Args) > 1 && (os.Args[1] == "__complete" || os.Args[1] == "__completeNoDesc")
 }
 
-func (g *GlobalFlags) GetLogger() *logrus.Logger {
+func (g *GlobalFlags) GetLogger() (*logrus.Logger, error) {
```
aufi, line 53: This seems to be the same as Ilanit's comment point 2:

GetLogger() caches g.logger before creating the audit file hook. When initConfig() initializes the logger from a --flags-file with an invalid audit-log path, it logs the error as a warning and leaves the partially initialized logger cached.
Later, during the same CLI invocation, Complete() calls GetLoggerOrDefault(). Since g.logger is already non-nil, GetLogger() returns (g.logger, nil), allowing the command to continue without an audit log.
Please only cache g.logger after audit.NewFileHook() succeeds, or persist and return the initialization error on subsequent calls.

aufi (changes requested):
Thanks Tammy, overall looks reasonable to me, I left a comment on fix logger caching in global_flags.go and asking for more tests (part likely in E2E).

Since we start make this audit log mandatory to run the CLI command, we need to be sure it works and handles corectly workarounds like pointing to /dev/null (details below), or edgecases including run on Windows OS. Real E2E tests could probably follow in another PR, but some additional testing shuld be part of this PR too.

One more point is not part of this PR, but it is closely related internal/audit/audit_logger.go:29 and its `chmod` operation. This is dangerous to do it always, e.g. unix-like approach with discarting audit log pointing it to /dev/null will fail on chmod. I'd suggest not run chmod, just create the file with those perms and rely/respect perms of already existing audit log paths.

main.go
```diff
@@ -33,6 +33,10 @@ func run() int {
 	}()
 	root := cobra.Command{
 		Use: filepath.Base(os.Args[0]),
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
aufi, line 36: This `PersistentPreRunE` initializes and caches the logger before `convert`\u0027s `Complete()` copies its local `--debug` value into `GlobalFlags.Debug`. As a result, the console hook is created with debug output disabled, and the subsequent `Debug = true` does not update its levels, so `crane convert --debug` no longer prints the debug messages emitted by `crane-lib/convert`. I confirmed this against the final PR commit with a runtime Cobra execution test. This should not block the PR because the deprecated `convert` command is planned for removal, but it is a real behavior regression worth recording.

aufi (commented):
Added comment for debug (affecting primary convert command, so no blocker, address as you prefer). I'm OK merging this PR after we cut branches for release-0.11, will approve in that point.

aufi (approved):

PR #961, opened 2026-09-10
:bug: accept encrypted section in comment

### LINKED ISSUE #900: [BUG] `--encrypt` rejects a valid rclone config when `[encrypted]` appears in a comment
**Summary**
When `[encrypted]` appears in a comment in a valid rclone config file, the file gets rejected .

Crane 0.11 Alpha on OCP

When `[encrypted]` appears in a comment in a valid rclone config file, the file shouldn't get rejected.

**What actually happened?**
When `[encrypted]` appears in a comment in a valid rclone config file, the file gets rejected .

**Steps to reproduce**
1. Create a source PVC `enctest` in `default` on the source cluster (any small
   RWO PVC).

2. Create a valid rclone config whose only occurrence of the token is a comment:

   ```console
      $ cat /tmp/c-comment.conf
   [remote]
   type = s3
   # note: do not add an [encrypted] section here
   ```

3. Run the transfer with `--encrypt`:

   ```console
   $ crane transfer-pvc \
       --source-context kind-source-cluster \
       --destination-context kind-dest-cluster \
       --pvc-name enctest:dest-a \
       --pvc-namespace default:default \
       --encrypt \
       --cloud-storage remote:bucket \
       --rclone-config-file /tmp/c-comment.conf

   [1/6] Reading source PVC ... ok
   [2/6] Creating destination PVC ... ok
   Error: rclone config already contains an [encrypted] section; remove it or omit --encrypt
   --- exit=1
   ```

There is no `[encrypted]` **section** in this config — only a comment mentioning
the word — yet crane refuses it.

stillalearner:
Fixes #900 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved encryption configuration validation to recognize only a genuine `[encrypted]` section.
  - Correctly handles section headers with trailing comments and configuration files with a leading UTF-8 marker.
  - Prevents false positives from comments, values, or similarly named sections.

- **Tests**
  - Added coverage for valid sections, encoded file markers, trailing comments, misleading text, and similarly named sections.


aufi (approved):
LGTM with note on possible edge case with UTF-8 "header" (discovered by AI), see below.

The new check misses an existing [encrypted] section when the file starts with a UTF-8 BOM (\ufeff[encrypted]), because strings.TrimSpace does not remove it. The previous substring check detected this case.

  Rclone recognizes the section and merges it with the generated one, retaining existing options such as filename_encryption = off. I reproduced this with rclone and confirmed that an isolated regression test fails against the new helper.

  Could we strip the leading BOM before checking section headers and optionally add a regression test.

PR #965, opened 2026-09-10
:bug: transfer-pvc[indirect]: fail early when rsync excludes rclone

### LINKED ISSUE #964: [BUG] Indirect transfer using rsync image without rclone  fails with an unclear error message
latest from main branch on KIND clusters

**Description**
If the rsync image predates rclone support , crane accepts the image, 
creates the destination PVC, launches the upload pod, and the container fails to start with 
`exec: "rclone": executable file not found in $PATH`.

crane **does** fail and returns a non-zero exit code like expected , but:

1. there is **no pre-flight check** that `--source-image` contains an `rclone`
   binary,
2. the top-level error **does not name the root cause** — it reports only
   `upload pod failed: pod <ns>/rclone-upload-<pvc> failed`, and
3. the failure is reached only **after** the destination PVC is created (a
   committed side effect → orphan), and only after the mover pod reaches a
   terminal state.

**What actually happened?**

nachandr:
Fixes [964](https://github.com/migtools/crane/issues/964)

## Bug Fixes

- Indirect cloud-storage transfers now validate that the selected transfer image includes a working `rclone` installation before creating the destination volume.
- Transfers fail earlier with clearer errors when the image cannot run `rclone`, helping identify invalid images or image-pull problems before work begins.
- Improved handling of transfer pod security settings during pre-transfer validation.
- Repeated or concurrent validation checks now use unique temporary probes, reducing conflicts between transfer attempts.

### DIFF
--- a/cmd/transfer-pvc/transfer-pvc.go
+++ b/cmd/transfer-pvc/transfer-pvc.go
@@ -1058,6 +1058,87 @@ func inspectPVCFileOwnership(c client.Client, namespace string, pvcName string, 
 	return &uid, nil
 }
 
+// verifyRcloneInImage runs a short-lived pod that execs "rclone --version" with
+// the given image and the mover pod's security context, confirming the image
+// ships a runnable rclone binary before crane commits any side effects.
+func verifyRcloneInImage(c client.Client, namespace string, image string, secCtx corev1.PodSecurityContext) error {
+	resolved := rsyncTransferImage(image)
+	podName := fmt.Sprintf("crane-rclone-check-%x", sha256.Sum256([]byte(resolved)))
+	if len(podName) > 63 {
+		podName = podName[:63]
+	}
+
+	nsName := types.NamespacedName{Name: podName, Namespace: namespace}
+	checkPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: namespace,
+		},
+		Spec: corev1.PodSpec{
+			RestartPolicy:   corev1.RestartPolicyNever,
+			SecurityContext: &secCtx,
+			Containers: []corev1.Container{
+				{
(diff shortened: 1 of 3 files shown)


cmd/transfer-pvc/transfer-pvc.go
```diff
@@ -1095,6 +1058,78 @@ func inspectPVCFileOwnership(c client.Client, namespace string, pvcName string,
 	return &uid, nil
 }
 
+// verifyRcloneInImage runs a short-lived pod that execs "rclone --version" with
+// the given image and the mover pod's security context, confirming the image
+// ships a runnable rclone binary before crane commits any side effects.
+func verifyRcloneInImage(c client.Client, namespace string, image string, secCtx corev1.PodSecurityContext) error {
```
aufi, line 1064: I'm sorry but I don't see reasonable starting a pod just to verify if the transfer image contains expected binary (which, as reported on the issue, fails with `exec: "rclone": executable file not found in $PATH` which looks descriptive enough).

Reasons:
- We ship the image and our E2E tests should fail if it was missing.
- If user adjusted image to some custom and even decided use indirect transfer, error on the pod looks more suitable to me, than start one more pre-validation pod on each and every pvc-transfer.

aufi (commented):
I'm open to discuss the issue in more holistic way, but as described below, starting always a pod to check presence of a binary we ship (and which should be validated by E2E test suite) looks as an overkill to me.

aufi (changes requested):
As mentioned in comment before I'm open to discuss the problem that issue for this PR is trying to solve (e.g. to ensure there are not orphan object if the failures happns directly on the pod) or better error message propagation to user.

But current implementation adding validation pod execution before each PVC transfer is -1 from me.

stillalearner:
I agree with @aufi , probably not needed.

PR #982, opened 2026-09-17
:bug: Support --audit-log /dev/null to disable audit logging

Tamar-Dinavetsky:
## Summary

Users can now pass `--audit-log /dev/null` to disable audit logging without encountering permission errors. This is a standard Unix idiom for discarding output.

## Problem

Previously, attempting to use `/dev/null` as the audit log path would fail with:
level=warning msg="Failed to open audit log file /dev/null: chmod /dev/null: operation not permitted"

This happened because `NewFileHook()` unconditionally called `Chmod(0600)` on all files, but device files like `/dev/null` don't support permission changes.

## Solution

- Check file type before applying `Chmod()` — only call it on regular files
- Create new files with mode `0600` directly instead of `0644`
- Skip `Chmod` on device files, allowing `/dev/null` to work as intended

## Changes

- **internal/audit/audit_logger.go**: Modified `NewFileHook()` to:
  1. Check if file exists with `os.Stat()`
  2. Only apply `Chmod(0600)` if it's a regular file (`IsRegular()`)
  3. Skip permission changes for device files

## Testing

- Verified `crane version --audit-log /dev/null` works without errors
- Verified default audit log still created and has correct permissions (0600)
- `/dev/null` remains unchanged after command execution

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audit log file handling for existing regular files.
  * Preserved appropriate error reporting when opening or updating audit log files fails.


internal/audit/audit_logger.go
```diff
@@ -26,10 +26,18 @@ func NewFileHook(path string, cmd *string) (*FileHook, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := f.Chmod(0600); err != nil {
+	fileInfo, err := f.Stat()
+	if err != nil {
 		_ = f.Close()
 		return nil, err
 	}
+	if fileInfo.Mode().IsRegular() {
```
aufi, line 34: This change fixes the originally reported problem 👍, but I'd suggest look on this issue from wider perspective.

The primary problem I can see is modification of the audit file attributes when it already existed. The tool might not know why the file has actual attributes like ownership or permissions. From my perspective the exected contract of the provided path for the audit log output (already existing) is only ability to write a data there, nothing else (to allow the smoothest possible functionality and flexibility).

If you have some time, please try analyze if it is more common&expected by our typical users, please do it, so we can discuss; otherwise I'd avoid the chown for already existing files fully.

Tamar-Dinavetsky: @aufi  good catch,
I went with only chmod'ing newly created files- existing ones keep their configured permissions. Device files skip chmod entirely. Feels like the safest default, but we can adjust if users need it.

